PR #159, opened 2014-02-08
Add g:javascript_indent_to_parens option

actionshrimp:
Added an option to allow customisation of indentation of function arguments. Rather than arguments being aligned with the opening function parenthesis (the default), arguments are indented one level extra. This works better with long function names and tabs.


bounceme:
this is now implemented, customizable like cindent

PR #173, opened 2014-03-14
highlight "use strict"; as PreProc

wadey:
This is part of the ECMAScript 5 standard, so I think it makes sense to
highlight it as a preprocessor command.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions_and_function_scope/Strict_mode

Updated to use backreference as suggested by @davidchambers


amadeus:
There's been no further interest in this PR and it can't currently be merged in anyways. Closing for now.

PR #199, opened 2014-07-08
Detect .es6 files as JavaScript

rbright:


bounceme:
great find!

bounceme:
I don't really get the point though

bounceme:
https://github.com/search?utf8=%E2%9C%93&q=if++extension%3Aes6&type=Code&ref=advsearch&l=&l=

i guess it is common enough

serv:
I think there are multiple ways to denote that a file uses es6 syntax.
I typically just use .js as the extension but it seems like there are standards that use .es6 as the file extension. In such cases, vim-javascript isn't detecting the file as javascript file.
Here's an example where some are using .es6 to denote use of es6.
https://github.com/emberjs/ember-rails/tree/master/app/assets/javascripts/ember-rails

PR #210, opened 2014-08-29
Update README.md with new Vundle-preferred syntax.

justinkim:


amadeus:
Closing because this was for `master`, and not `develop`. It has been replaced with #230

PR #213, opened 2014-09-21
Add @api annotation syntax highlighting

jedborovik:
I know the standard jsDoc uses @public and @private to mark visibility but @api public or @api private seems reasonably common in javascript files (like most of TJ Holowaychuk's repos). This pull request adds syntax highlighting for @api tag


amadeus:
Looks good to me!

PR #228, opened 2015-01-07
Change jsPrens, brackets, etc. to be colorized

isaachess:
Change the `HiLink` for several js syntax items so that they are colorized more commonly by color schemes.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -268,11 +268,11 @@ if version >= 508 || !exists("did_javascript_syn_inits")
   HiLink jsBooleanTrue          Boolean
   HiLink jsBooleanFalse         Boolean
   HiLink jsNoise                Noise
-  HiLink jsBrackets             Noise
-  HiLink jsParens               Noise
-  HiLink jsBraces               Noise
-  HiLink jsFuncBraces           Noise
-  HiLink jsFuncParens           Noise
+  HiLink jsBrackets             Function
+  HiLink jsFuncBraces           Function
+  HiLink jsBraces               Function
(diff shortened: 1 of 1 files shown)


qstrahl:
Howdy! Two things:

First, please make your pull requests against the `develop` branch; that's where active development takes place.

Second, your choice of syntax groups seems odd to me. In `:h syntax`, the `Structure` group is paired with the examples `struct, union, enum, etc.` This does not seem like a suitable category for parens. Can you find examples of popular syntax files with highlights like this?

amadeus:
:+1: to what @qstrahl said.

And I would add, while it might make sense for `jsFuncBraces` to be part of Function, it does not make sense for `jsBrackets` and `jsBraces`.

Also, if this is something that looks good for your particular theme, I suggest you create this simple fix for yourself. Create the following file `after/syntax/javascript.vim` and add these lines to it:

```
hi! def link  jsBrackets             Function
hi! def link  jsFuncBraces           Function
hi! def link  jsBraces               Function
hi! def link  jsParens               Structure
hi! def link  jsFuncParens           Structure
```

Then you can easily override our groups and use your own without us needing to make the semantic change, which, as I agreed with @qstrahl, isn't quite accurate.

Of course, you can also just add whatever you want into that file I just suggested, to assign your own arbitrary colors to our specific groups:

```
hi jsFuncBraces            guifg=#cefdff   guibg=#000000   gui=NONE
```

These will override any colors they may have inherited from being linked.

Anyways, closing this for now. Obviously the discussion can continue, but anything we'd consider would need to be a PR against the `develop` branch.

PR #229, opened 2015-01-08
Disable custom gq for comments

### LINKED ISSUE #226: vim-javascript no longer lets gq breaks long comments
Previous to commit ca201905da0b44b703cd744e65d29bb9d3a11926 vim-javascript would not interfere with `gq`'s breaking of comments that were longer than textwidth into multiple lines, so for example, prior to the commit this works:

```
:set textwidth=20
// a comment longer than 20 characters
shift-v
gq
// a comment longer
// than 20
// characters
```

but after ca201905da0b44b703cd744e65d29bb9d3a11926 `gq` has no effect on the first example, and on the second example `gq` actually merges the three lines into one (overflowing the textwidth).

JelteF:
Fixes #226

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -443,8 +443,14 @@ function! Fixedgq(lnum, count)
     let l:tw = &tw ? &tw : 80;
 
     let l:count = a:count
+    let l:first_char = indent(a:lnum) + 1
 
     if mode() == 'i' " gq was not pressed, but tw was set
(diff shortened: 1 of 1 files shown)


amadeus:
Looks good to me, given that this issue has been pretty active, I'll merge it in shortly unless there are any objections.

davidchambers:
LGTM

PR #232, opened 2015-01-15
Fixed block comments and sectional alignment issues

amadeus:
I found both a bug, and what I deemed a bit of semantic issue with block comments in JS.

First off, we have the syntax region called `jsDocComment`, which is what catches block comments. However, it had an extra specifier requiring two asterisks, which could lead to issues like this:

example

I have done the following - removed the double asterisk requirement, and renamed the group to be `jsBlockComment`.

All the JS Doc highlighting should still work as expected, it's just the containing group has been renamed, and fixed.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -79,26 +79,26 @@ endif   "" JSDoc end
 syntax case match
 
 "" Syntax in the JavaScript code
-syntax match   jsFuncCall        /\k\+\%(\s*(\)\@=/
-syntax match   jsSpecial         "\v\\%(0|\\x\x\{2\}\|\\u\x\{4\}\|\c[A-Z]|.)" contained
-syntax match   jsTemplateVar     "\${.\{-}}" contained
-syntax region  jsStringD         start=+"+  skip=+\\\("\|$\)+  end=+"\|$+  contains=jsSpecial,@htmlPreproc,@Spell
-syntax region  jsStringS         start=+'+  skip=+\\\('\|$\)+  end=+'\|$+  contains=jsSpecial,@htmlPreproc,@Spell
-syntax region  jsTemplateString  start=+`+  skip=+\\\(`\|$\)+  end=+`\|$+  contains=jsTemplateVar,jsSpecial,@htmlPreproc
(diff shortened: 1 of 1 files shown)


syntax/javascript.vim
```diff
@@ -225,7 +225,7 @@ if version >= 508 || !exists("did_javascript_syn_inits")
   HiLink jsComment              Comment
   HiLink jsLineComment          Comment
   HiLink jsEnvComment           PreProc
-  HiLink jsDocComment           Comment
+  HiLink jsBlockComment           Comment
```
davidchambers: :triangular_ruler: It would be nice to delete two spaces to preserve column alignment here and on `:56`.

amadeus: Good call, I'll fix this, was in a hurry xD

amadeus:
Fixed!

syntax/javascript.vim
```diff
@@ -53,7 +53,7 @@ if !exists("javascript_ignore_javaScriptdoc")
   "syntax include @javaHtml <sfile>:p:h/html.vim
   "unlet b:current_syntax
 
-  syntax region jsDocComment      matchgroup=jsComment start="/\*\*\s*"  end="\*/" contains=jsDocTags,jsCommentTodo,jsCvsTag,@jsHtml,@Spell fold
+  syntax region jsBlockComment      matchgroup=jsComment start="/\*\s*"  end="\*/" contains=jsDocTags,jsCommentTodo,jsCvsTag,@jsHtml,@Spell fold
```
davidchambers: :triangular_ruler: here too? It's not obvious whether `matchgroup` is meant to align with something so it doesn't really matter.

amadeus:
Durp. Fixed again! It looks off when you look at all the items in place after it.

amadeus:
There are some other general alignment issues, that I could fix in other places too, for consistencies sake.

amadeus:
Added another commit that fixed some unrelated alignment issues. Also found a couple deprecated comments near the bottom that I removed as well.

davidchambers:
LGTM

PR #233, opened 2015-01-21
Adds better es6 support

goatslacker:
This is an all-around improvement for ES6 syntax highlighting PR:

* Removes CommonJS highlighting because it is not spec compliant.
* Adds ES6 modules syntax.
* Moves `class` and `extends` from jsKeyword into Structure.
* Makes template strings support multiline.
* Removes `static` and `super` from reserved words.
* Adds tagged template strings.
* Adds `WeakMap`, `Set`, `Map`, `Symbol`, and `Promise` to globals.
* Adds `of` keyword.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ variables:
     let g:javascript_conceal_undefined  = "¿"
     let g:javascript_conceal_NaN        = "ℕ"
     let g:javascript_conceal_prototype  = "¶"
+    let g:javascript_conceal_static     = "•"
+    let g:javascript_conceal_super      = "Ω"
 
 ## Contributing
 
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -35,7 +35,11 @@ syntax keyword jsOperator       delete instanceof typeof void new in
 syntax match   jsOperator       /\(!\||\|&\|+\|-\|<\|>\|=\|%\|\/\|*\|\~\|\^\)/
 syntax keyword jsBooleanTrue    true
 syntax keyword jsBooleanFalse   false
-syntax keyword jsModules        import export default from as
+syntax match jsModules          /import\|export/
+syntax keyword jsModuleWords    default from as contained
+
+syntax region jsModuleContainer      start="\(import\|export\) \?" end="$" contains=jsModuleWords,jsStringS,jsStringD,jsTemplateString,jsModules,jsArrowFunction,jsClass,jsFunction,jsStorageClass,jsOperator
+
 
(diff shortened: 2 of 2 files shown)


amadeus:
Looks good to me.

Is there anything special we should incorporate for the new class API? As in, properly matching it's semantic structure?

goatslacker:
Not unless we want to highlight method names or the constructor. Related to #218

amadeus:
I have not really looked into ES6, so I can't fully comment on the meaning of the various portions of the syntax, however if you compare it in my theme, which takes advantage of some of the special function declarations, you'll see we are not recognizing class definitions as function definitions

example

(Also, I don't think this PR needs to be delayed for merging because of this, just food for thought as something to add)

goatslacker:
Another point to think about is we probably shouldn't highlight `from` and `as` unless they're in the context of `import`

Same goes for `default` unless it's in the context of `export`

goatslacker:
yeah, again this adds just basic support.

Would be cool not to highlight:

```js
var from = 1
var as = 2
```

amadeus:
Do you want to do that before we merge in? Otherwise, I think this PR is good to go in.

goatslacker:
Alright this is good to go

screen shot 2015-01-25 at 12 42 01 pm
screen shot 2015-01-25 at 12 42 11 pm

goatslacker:
There's one small issue that would be difficult to address. We need an object literal syntax region `jsBlock` isn't good enough.

goatslacker:
Tagged template strings

screen shot 2015-01-25 at 1 45 02 pm

README.md
```diff
@@ -74,6 +74,8 @@ variables:
     let g:javascript_conceal_undefined  = "¿"
     let g:javascript_conceal_NaN        = "ℕ"
     let g:javascript_conceal_prototype  = "¶"
+    let g:javascript_conceal_static     = "•"
+    let g:javascript_conceal_super      = "Ω"
```
goatslacker: feel free to bikeshed these

goatslacker:
`of` highlighting

screen shot 2015-01-25 at 2 00 25 pm

goatslacker:
Default args. Should we also highlight what's inside?

screen shot 2015-01-25 at 2 15 36 pm

davidchambers:
> Default args. Should we also highlight what's inside?

I think so, if possible. :)

goatslacker:
Cool that can go in a separate PR I think I need to refactor some of the syntax

amadeus:
:+1:

PR #236, opened 2015-01-28
Fixes module syntax for common js

goatslacker:
It's currently broken if you do things like

```js
exports.foo = 2
```


amadeus:
:+1:

Issue #34, opened 2011-11-21
multiline var statements not indented


wmertens:
I'm trying to follow the rules that JSLint has, and something like

```javascript
var a=b,
    c,
    d;
```

gets indented as

```javascript
var a=b,
c,
d;
```

which JSLint doesn't like.

I may also be doing something wrong of course.

kmckee:
+1

amadeus:
Should be fixed in `develop`

PR #238, opened 2015-01-29
Fix for #34

dragn:
Fix for #34 
Regex modified to detect lines, ended with comma and containing single `=` (equals) character (not double or triple to avoid match on conditionals).
May break things if you do assignments in weird places, for example: 
```js
A = {
    B : (C = D),
        E : F,
};
```


davidchambers:
LGTM. Could you squash the commits?

dragn:
Done

PR #239, opened 2015-02-02
Adds jsClass to all expressions

goatslacker:
This was missing before. Now class highlights properly inside other blocks.


davidchambers:
LGTM

Issue #139, opened 2013-10-10
dot chained calls are not indented correctly


nisaacson:
When chaining calls together via the `.` operator, the following lines are not indented correctly

# Desired
```javascript
one()
  .then(two)
  .then(function() {
  })
  .fail(failHandler)
```

# Current
```javascript
one()
.then(two)
.then(function() {
})
.fail(failHandler)
```

Is there a way I can add a custom regular expression to perform additional indentation after the standard vim-javascript indent plugin has finished? Adding an additional two spaces to any line that begins with `.` would be a vast improvement

sukima:
I don't think there is any reliable way to do this and not force semi-colons (without tricks like syntax). Now for me that's totally okay since I like semi-colons (for this very reason) but many developers prefer the more renegade style of flexing JavaScript's [auto semi-colon insertion (ASI)](http://jamesallardice.com/understanding-automatic-semi-colon-insertion-in-javascript/).

Thoughts?

goatslacker:
If it starts with a dot, indent. This is probably not bullet proof however.

sukima:
@goatslacker good point. I guess I took for granted that the dot can also be trailing. Investigating...

goatslacker:
So I did this for leading commas and it worked for 80% of the use cases, not sure if dot would be the same case or not. I haven't fully thought this through.

Trailing dot wouldn't work, but I guess if the dot is trailing you could also indent. I think this is the current behaviour actually.

sukima:
@goatslacker see PR #248. My dev tests show this as working.

devm33:
Any update on this? I'd be happy to spend some time getting this working.

sukima:
@devm33 No progress, I've learned to live with the dot as a line prefix not a suffix.

lvarayut:
Any progress on this issue? I'm having the same problem.

chalmagean:
+1

Skinney:
+1

randunel:
Had to build my own fix for the dot start https://github.com/randunel/vim-javascript. Would be nice to include something similar in this lib though.

djmccormick:
+1

ksmithbaylor:
@randunel Is that something you could make a pull request out of? How well does it work?

mistercrunch:
+1

bounceme:
a problem no longer

jeffrom:
A hero in our time

floscr:
[...]

PR #248, opened 2015-03-03
Add dot-chaining syntax to main indent logic

sukima:
Most style guides depict chained methods to have a leading dot ans be indented on level deeper then the initial statement.

This patch adds a period as a key symbol for a nested indent. In the same fashion as the leading comma does.

It allows an indent of the following:

```javascript
var x = {
  x: 'foo',
  y: 'bar',
  z: (function() {
    return [1, 2, 3]
      .map(toFruits)
      .compact()
      .sortBy('color');
  })()
};
```

This should close issue #139.

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -303,7 +303,7 @@ function GetJavascriptIndent()
   " If we got a closing bracket on an empty line, find its match and indent
   " according to it.  For parentheses we indent to its column - 1, for the
   " others we indent to the containing line's MSL's level.  Return -1 if fail.
-  let col = matchend(line, '^\s*[],})]')
+  let col = matchend(line, '^\s*[],.})]')
   if col > 0 && !s:IsInStringOrComment(v:lnum, col)
     call cursor(v:lnum, col)
 


goatslacker:
Cool. Let me load this one and live with it for a while and see if anything breaks.

goatslacker:
There's a little more work that needs to go into this one. While this doesn't re-indent chained dots it doesn't automatically indent them either

sukima:
Crumbs, I'll look into it.

sukima:
Yeah, looking into this the fix won't help. Closing this and will make a new pull request later. Bummed out but glad you checked it.

goatslacker:
It may be worthwhile checking this in regardless if it doesn't screw with anything else. Because the re-indenting of dot chain calls is super annoying.

sukima:
@goatslacker up to you. I don't mind. I'm still going to hunt down a proper solution eventually.

ksmithbaylor:
Any update on this? I'd love to see it move forward and would be willing to help out.

sukima:
@ksmithbaylor no update. I just learned to live with it.

Issue #195, opened 2014-06-16
if statements in switch case


u9520107:
currently i see switch cases to be indented as such. Not sure where I did wrong. 
````
switch(name) {
    case 1:
        break;  //doesn't align with case
    case 2:
        a = b;   //seems normal if there's a statement somewhere
    break;
    case 3:
        if (true) {
    }   //a if block breaks the indentation
    c = d;
    break;
}
````

carlsverre:
I am seeing the same issues, has anyone sorted out a solution?

virgilio:
+1

PR #249, opened 2015-03-03
Manage case indent

utix:
* Multi case
* half indent for case
* indent when typing : in case

case is now use as a block delimiter
#195 should be fixed

Before:

````javascript
switch(name) {
    case 1:
        break;
    case 2:
        a = b;
    foo();
    bar();
    break;
    case 4:
        case 3:
        if (true) {
        foo();
    }
    bar();
    c = d;
    break;
    default:
        foo();
    break;
}
````
After:

````javascript
switch(name) {
  case 1:
    break;
  case 2:
    a = b;
    foo();
    bar();
    break;
  case 4:
  case 3:
    if (true) {
        foo();
    }
    bar();
    c = d;
    break;
  default:
    foo();
    break;
}
````

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -68,6 +68,16 @@ let s:comma_last = ',\s*$'
 let s:ternary = '^\s\+[?|:]'
 let s:ternary_q = '^\s\+?'
 
+let s:case_indent = &sw
+let s:case_indent_after = &sw
+let m = matchlist(&cinoptions, ':\(.\)')
+if (len(m) > 2)
+    let s:case_indent = m[1]
+endif
+let m = matchlist(&cinoptions, '=\(.\)')
+if (len(m) > 2)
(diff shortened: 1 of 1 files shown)


goatslacker:
Oh wow this is hot. Let me load it up and play with it.

amadeus:
I am personally not a fan of the half indent stuff, I like to use full indents always. Is there a way this could be turned on or off?

utix:
I use `set cinoptions+=:2,=2` (Too many C I think :) )
I will look to use this param, but I think my patch isn't worst than the actual indent of multilines case. If someone says how to retrieve those values just say me I will do the patch.

@amadeus  you can remove the `/2`on lines 308 310 406

goatslacker:
One more nit, can you send this PR to the develop branch?

utix:
Sure. I have found how to get the value, it will be configurable

PR #260, opened 2015-04-03
Add Node.js specific globals

nickleefly:


goatslacker:
These aren't JavaScript specific syntax so they don't belong in a JS syntax highlighter. With ES6 we have module syntax which is already being highlighted.

nickleefly:
ok, thanks

PR #264, opened 2015-04-26
Add `const` and `let` to indent file

lencioni:
These keywords are part of ES6 and are used like the `var` keyword. This commit makes the indentation file treat them the same as `var`.

To be honest, I was unable to tell if this had any effect on the behavior. However, it seems like this is an appropriate change to make for completeness. I'd love your feedback on whether you think this makes sense or not.


goatslacker:
Can you open the PR against the develop branch?

PR #266, opened 2015-04-26
Add `const` and `let` to indent file

lencioni:
These keywords are part of ES6 and are used like the `var` keyword. This
commit makes the indentation file treat them the same as `var`.


qstrahl:
Hey, @lencioni, thanks for the patch! :D

Issue #107, opened 2013-08-07
`arguments` keyword not consistently highlighted


qstrahl:
Example code:
```javascript
function MyDate () {
    Date.apply(this, arguments); // arguments is not highlighted here
}
```

It should be highlighted *anywhere* in a function block.

crusoexia:
global keyword + 1
It is very weird for people to use 'arguments' out of function

qstrahl:
I think I'm favour of that, too. It's easy enough and I doubt anyone would be complaining about it.

goatslacker:
+1

qstrahl:
Closed by #268

PR #268, opened 2015-05-28
fix issue #107, make 'arguments' become global keyword

crusoexia:
after the fix:
arguments


qstrahl:
Looks good to me! Let's give 'er a whirl.

PR #274, opened 2015-07-15
fix: export region should support es6 class and block openings

ccarpita:
Currently, the following syntax results in an error on the closing
brace:

```js
export default {
} // <-- error here

export default class Modules extends Base {
} // <-- error here
```

This patch adds jsBlock to be included in a jsExportContainer region
so that region matching will continue to the end of the class or block.


qstrahl:
Looks good, but please open the request against the `develop` branch

Issue #256, opened 2015-03-18
export class MyClass { ... highlights closing brace as error (es6)


euoia:
When using the following ES6 syntax:
```
export class MyClass {
...
}
```

The closing brace is highlighted as an error. See screenshot.

![image](https://cloud.githubusercontent.com/assets/1271216/6701560/cb7ce908-ccf4-11e4-8246-431ada8747a9.png)

Description of the export statement: http://www.2ality.com/2014/09/es6-modules-final.html.

nemtsov:
Same thing with exporting objects:

![export_objects_vim_bug](https://cloud.githubusercontent.com/assets/39958/8366100/dd21293a-1b5f-11e5-9ad8-06d5065b9c4f.png)

gf3:
being highlighted as a `jsParensError`

SirCmpwn:
Does not seem to have been fixed by #275 so far as I can tell.

ccarpita:
@SirCmpwn are you sure you're on develop branch?  Otherwise please post some sample code that generates incorrect syntax and I can test locally.

SirCmpwn:
Oh, whoops, didn't realize this was only in develop.

nemtsov:
#275 resolves the issue for me.

gf3:
it didn't for me, tried it a couple days ago

ccarpita:
@gf3 can you post a test case that fails?  Also, sorry to ask, but are you 100% sure you're using the current `develop` branch?

toddw:
Latest `develop` branch resolved the problem for me

gf3:
previous issue seems to be fixed since i last tried, looking good!

PR #275, opened 2015-07-17
fix: export region should support es6 class and block openings

ccarpita:
Currently, the following syntax results in an error on the closing
brace:

```js
export default {
} // <-- error here

export default class Modules extends Base {
} // <-- error here
```

This patch adds jsBlock to be included in a jsExportContainer region
so that region matching will continue to the end of the class or block.


ccarpita:
This commit fixes issue GH-256

qstrahl:
Let's give this a shot

goatslacker:
I remember this breaking certain things, lets give it a whirl now.

Issue #272, opened 2015-06-19
Arrow function only works if surrounded by parentheses


rightaway:
The arrow gets highlighted fine in this case
```
(arg) => {}
```

But not when the parentheses are omitted.
```
arg => {}
```

davidchambers:
Do we close an issue when the relevant PR is accepted, or once the commit is pulled into master?

qstrahl:
Eh, let's close them sooner to avoid forgetting they exist.

PR #276, opened 2015-07-17
fix: Highlight arrow functions with bare single params

ccarpita:
Arrow expressions with "naked" single parameters are currently not being
recognized as such, due to the mis-labeling of the expression as assignment.

```js
// The part: `x =` looks like assignment to syntax rules:
let myFunc = x => 3;
```

This change introduces a negative lookahead for the `>` character in
the assignment matcher.

Issue: GH-272


qstrahl:
*Slightly* nervous about the lookahead, but let's try it and see if anyone reports bad behaviour / slowness

davidchambers:
Thanks, @ccarpita!

ccarpita:
Good point about the lookahead, though it's single char and should evaluate quickly for assignment expressions.  If the performance suffers we should probably create a region type for the full arrow expressions with prioritized matching over assignment

Issue #247, opened 2015-03-01
  Invalid syntax highlighting for single-line comments after an import statement


chocolateboy:
vim-javascript: 0.10.0 (e9709a9fb29ed0739eb244b5e70211c1ab647535)
vim: 7.4.52
OS: Linux (Ubuntu 14.04.2)

```javascript
import foo from 'bar'; // this isn't highlighted correctly
import baz from 'quux'; /* this is OK */
```

On my system, the comment on the first line is initially unhighlighted (white) until the single quote, which throws it off, and the rest of the line is highlighted like a string.

The second line is highlighted correctly.

PR #277, opened 2015-07-17
fix GH-247: Single-line comments after import statement

ccarpita:
The single line comment in the following statement is currently no
highlighted:

```js
import Y from 'Y' // this should be a comment
```

This introduces jsLineComment as being containable in an import statement,
and also closes the import statement on semicolon so that other expressions
can be highlighted correctly.


ccarpita:
Fixes GH-247

qstrahl:
Seems harmless

PR #279, opened 2015-07-17
Adds support for ES7/ES2016 decorators

jedverity:
See here for an explanation of decorators: https://github.com/wycats/javascript-decorators

I'm a Vim syntax newb so forgive errors. Thanks to @kergoth for this inspiration: https://github.com/kergoth/dotvim/blob/master/bundle/hi-py-decorators/after/syntax/python/decorators.vim


goatslacker:
Awesome, thanks

PR #286, opened 2015-08-10
Allow customization of line continuation indent

bl-nero:
This change introduces a `g:javascript_continuation_indent` variable, which works similarly to the "+N" component of `cinoptions`. The default value is `&sw`, and that means that the default behavior remains unchanged.


bl-nero:
Sorry, wrong branch. :/ Closing.

PR #288, opened 2015-08-18
Add support for ngdocs

whitneyit:
Added support for [angular's](https://angularjs.org) own [ngdoc](https://github.com/angular/angular.js/wiki/Writing-AngularJS-Documentation) syntax.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -63,7 +63,7 @@ if !exists("javascript_ignore_javaScriptdoc")
   syntax region jsBlockComment    matchgroup=jsComment start="/\*\s*"  end="\*/" contains=jsDocTags,jsCommentTodo,jsCvsTag,@jsHtml,@Spell fold
 
   " tags containing a param
-  syntax match  jsDocTags         contained "@\(alias\|api\|augments\|borrows\|class\|constructs\|default\|defaultvalue\|emits\|exception\|exports\|extends\|file\|fires\|kind\|listens\|member\|member[oO]f\|mixes\|module\|name\|namespace\|requires\|template\|throws\|var\|variation\|version\)\>" nextgroup=jsDocParam skipwhite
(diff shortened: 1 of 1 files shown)


Delapouite:
I would prefer this syntax highlighter to remain framework specificities free.

qstrahl:
I'm actually in favour of this; this is unlikely to affect anyone not using
angular, and would be nice to have for angular users

qstrahl:
I'm going to need to see some more opinions on this before this can be merged.

amadeus:
So, personally I am not a fan of adding these to the syntax file, since I think it will add a lot of mess, and non-standard stuff.

However, what I personally wouldn't mind, is if we put these types of plugin/library specific stuff in special individual files, and then added some logic in the main syntax file to source those files if a particular global variable is set.

Then we could encapsulate that special highlighting separately, and allow new libs to be added if people wanted them without much damage/mess added to the main syntax file. Thoughts?

qstrahl:
That sounds like the best of both worlds; I'm for it.

whitneyit:
I'm all for having framework/plugin specific annotations 

That would then allow more specific highlighting definitions.

For instance the `link` rule isn't working 100%.

It needs to be able to work with the following:

> {@link foo.bar.baz.method Link Title}

Currently the braces and title are not being highlighted

davidchambers:
Nice suggestion, @amadeus! I hope it's feasible.

amadeus:
I have an idea for how this could work, let met think about it over this weekend and I'll see if I can get something drawn up.

whitneyit:
I've rebased these changes off of develop. @amadeus is there anything that I can do to help get this PR merged?

whitneyit:
@amadeus did you come up with a way to define custom plugin/library/framework syntax files?

qstrahl:
@whitneyit Please update this PR and let's see if we can't get it merged in.

whitneyit:
@qstrahl Updated!

amadeus:
I have added support for this in #470

Issue #291, opened 2015-08-28
indent after one-line if statement without semicolon


camsteffen:
This is incorrectly indented

    if(1) return
        console.log('foo')

qstrahl:
Oh man, I have a feeling this one is going to be absurdly difficult to solve. Thanks for the report.

cambunctious:
Admittedly I am not familiar with the code, but shouldn't this be pretty simple? If there is anything after the parenthesis, then that is the if block and the next line is not.

cambunctious:
I decided to get my hands dirty.

Looking at `indent/javascript.vim`...

The problem is that `s:InOneLineScope(lnum)` evaluates true (non-zero) with the line `if(1) return`. I believe the function would be better named `BeginsOneLineScope`. Fixing it will require changing `s:one_line_scope_regex` so that it does not match when there is a statement following the `if/for/while/...`. I'm not very good at Vim regular expressions but I will keep at it unless someone wants to step in.

cambunctious:
Related to #192 and #253

qstrahl:
Thanks for the detective work, @cambunctious!

Using regexes for this stuff makes me throw up in my mouth a bit, but unless/until we can make use of the underlying syntax groups to do this stuff, it's pretty unavoidable... I have a knack for vim regex so I'll take a crack at getting something in place that won't break under the smallest amount of pressure, and hopefully we'll be able to put this to bed for now.

qstrahl:
Also, with regard to:

> Admittedly I am not familiar with the code, but shouldn't this be pretty simple? If there is anything after the parenthesis, then that is the if block and the next line is not.

Corner cases. Consider the following:

```javascript
if (something) { doLittle(); } else
    bigNamedVariable.doTonsOfStuffHolyCrapThisMethodNameIsSoConvenientlyDescriptive();
```

I don't have a very good idea of how all those indent-related functions operate, but I assume that cases like the above could throw it for a loop. We can ignore those cases, but only until somebody posts an issue here complaining that it doesn't work. :P

cambunctious:
Maybe `s:InOneLineScope(lnum)` could work something like this:

1. if the line ends with `else`, return true
1. If the line ends with `if|while|for` followed by a parenthetical, return true
1. return false

I see the tricky part is the parenthetical with regex. Is it possible to use the `%` command to find matching parenthesis? If so, start step 2 by detecting a `)` at the end of the line.

qstrahl:
[...]

PR #293, opened 2015-09-07
Issue #291 better one line scope indentation

camsteffen:


qstrahl:
@camsteffen Can you elaborate on what this change addresses? Also you'll need to rebase if you want it merged in

davidchambers:
Presumably this solves the issue raised in #291:

>     if(1) return
>         console.log('foo')

qstrahl:
@camsteffen It looks like you merged our `develop` branch into yours; you will have to undo that change with a `git reset --hard b246a506bd1bdfa0827a9861707ffb0b6934b3c5` and instead do a rebase with `git rebase refs/remotes/pangloss/develop` You will have to resolve the resultant merge conflict.

camsteffen:
Thanks for helping with the rebase. Hopefully it is set to go now.

To recap the change, the regex now checks to see if there is a statement after the parenthesis for a `if|for|while` so that `if(foo) bar()` does not begin a one line scope. If a line ends with `else`, then it does begin a one line scope. This pattern could potentially break for something like `while("(foo)" === bar)` but that is an edge case.

qstrahl:
I'm willing to try it out for now. We can revert if breakages come up.

bounceme:
hey, some things are still being indented
```
function n (err) {
  if (err) return log(err)
    var act = 0
}
```

camsteffen:
Unfortunately I think we have the best solution possible using regular expressions. It becomes an ultimatum between (1) falsely indent after `if(foo) bar()` or (2) falsely not indent after `if(foo())`. I prefer option 1, especially since it can be fixed with a semicolon: `if(foo) bar();`

indent/javascript.vim
```diff
@@ -66,7 +66,7 @@ let s:continuation_regex = '\%([\\*+/.:]\|\%(<%\)\@<![=-]\|\W[|&?]\|||\|&&\|[^=]
 " TODO: this needs to deal with if ...: and so on
 let s:msl_regex = s:continuation_regex.'|'.s:expr_case
 
-let s:one_line_scope_regex = '\%(\<\%(if\|else\|for\|while\)\>\|=>\)[^{;]*' . s:line_term
+let s:one_line_scope_regex = '\%(\<else\>\|\<\%(if\|for\|while\)\>\s*(.*)\)' . s:line_term
```
bounceme: unexpectedly this seems to be working:
```
let s:one_line_scope_regex = '^[})\]]*\s*\(if\|else\|while\|try\|catch\|finally\|for\|else\s\+if\)\s*_\=$'
```
gotten from https://github.com/jiangmiao/simple-javascript-indenter/blob/master/indent/javascript.vim#L350

bounceme:
never mind, there isn't indentation when the curly braces are omitted

camsteffen:
can you give an example?

bounceme:
```
  if (true)
  console.log("value");
```

camsteffen:
Your example indents correctly for me...

bounceme:
did you replace line 69 with
```
let s:one_line_scope_regex = '^[})\]]*\s*\(if\|else\|while\|try\|catch\|finally\|for\|else\s\+if\)\s*_\=$'
```

camsteffen:
No, but now I'm confused. Why are we looking at this other line of code if it doesn't work?

bounceme:
the indent script that it is copied from, jiangiao's, does correctly indent these different variations of ifs.maybe we can get a similar behavior by looking at other parts of that script to see what he did

camsteffen:
Sorry, it doesn't solve the problem that I described before:
> It becomes an ultimatum between (1) falsely indent after `if(foo) bar()` or (2) falsely not indent after `if(foo())`. I prefer option 1, especially since it can be fixed with a semicolon: `if(foo) bar();`

Here's another way to look at it. Using regular expressions, there is no way to detect whether parenthesis are matching. It can't see if `if (` matches with `)` at the end of the line.

bounceme:
well, it may be possible if we try with ```:h searchpair()```

camsteffen:
Good find, I didn't know about that. Maybe I'll look into it when I have some time. It would be more than a one-liner fix.

bounceme:
This is what I think would be easiest,though not elegant:
```
let s:one_line_scope_regex = '\%(\%(\<else\>\|\<\%(if\|for\|while\)\>\s*(\%([^()]\|[^()]*([^()]*)[^()]*\))\)\|=>\)' . s:line_term
```
This can be extended to allow nesting more than two pairs

bounceme:
last comment showed two pairs nested,here's what 4 looks like...
```
let s:one_line_scope_regex = '\%(\%(\<else\>\|\<\%(if\|for\|while\)\>\s*(\%([^()]*\|[^()]*(\%([^()]*\|[^()]*(\%([^()]*\|[^()]*([^()]*)[^()]*\))[^()]*\))[^()]*\))\)\|=>\)' . s:line_term
```

PR #312, opened 2015-10-06
Add common DOM Functions

simono:
getElementById and getElementsByClassName are two often used Functions.


Delapouite:
Thanks for this addition @simono 
But what about other popular functions like `querySelector` or `querySelectorAll`… ?

I see two exclusive options here : 

- extract all the DOM / HTML specific syntax to its own companion plugin. By developping it independently, it will be more complete. Also, people working with pure JS on the server side will have a lighter plugin without all the client side mess.
- be exhaustive and follow the spec, which is quite huge.

davidchambers:
> extract all the DOM / HTML specific syntax to its own companion plugin

I like this idea.

Although I'm against vim-javascript giving special attention to DOM APIs, since we already do so it would be completely reasonable to accept this patch (though we'd need a new pull request against `develop`).

simono:
@davidchambers @Delapouite ebc8386 is against `develop` and I've add `querySelector` and `querySelectorAll`.

Adding all DOM Functions this way is indeed tedious…

davidchambers:
> ebc8386 is against `develop` and I've add `querySelector` and `querySelectorAll`.

Great. GitHub doesn't let one change a pull request's source or destination branch, so you'll need to open a new pull request.

Issue #332, opened 2015-12-10
Multi-line arrow functions


ksmithbaylor:
Currently, when the body of an arrow function that returns an immediate value is split over multiple lines, the second line (the value) is aligned even with the line above:

```javascript
const someVeryLongIdentifier = (a, b, c) =>
anotherVeryLongExpressionThatWillNotFitOnASingleLineNicely();

nextLine();
```

In my opinion, it should be indented one level more than the line on which it started, but still allow the following lines to stay at the same level as before:

```javascript
const someVeryLongIdentifier = (a, b, c) =>
  anotherVeryLongExpressionThatWillNotFitOnASingleLineNicely();

nextLine();
```

This should also apply to arrow functions defined as function arguments. Current behavior:

```javascript
someVeryLongIdentifier.getThings().filter(Boolean).map(x =>
                                                       anotherVeryLongExpressionThatWillNotFitOnASingleLineNicely());

                                                       nextLine();
```

Expected behavior:

```javascript
someVeryLongIdentifier.getThings().filter(Boolean).map(x =>
  anotherVeryLongExpressionThatWillNotFitOnASingleLineNicely());

nextLine();
```

WChargin:
The following patch works for me. Can someone familiar with the repo make sure this seems reasonable? If it does, I'll pull-request it.

```diff
diff --git a/indent/javascript.vim b/indent/javascript.vim
index a935a87..f1329cb 100644
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -55,7 +55,7 @@ let s:continuation_regex = '\%([\\*+/.:]\|\%(<%\)\@<![=-]\|\W[|&?]\|||\|&&\|[^=]
 " TODO: this needs to deal with if ...: and so on
 let s:msl_regex = s:continuation_regex.'|'.s:expr_case
 
-let s:one_line_scope_regex = '\<\%(if\|else\|for\|while\)\>[^{;]*' . s:line_term
+let s:one_line_scope_regex = '\%(\<\%(if\|else\|for\|while\)\>\|=>\)[^{;]*' . s:line_term
 
 " Regex that defines blocks.
 let s:block_regex = '\%([{[]\)\s*\%(|\%([*@]\=\h\w*,\=\s*\)\%(,\s*[*@]\=\h\w*\)*|\)\=' . s:line_term

```

[...]

Issue #338, opened 2015-12-30
Please add an explicit license file to the repo


mbrukman:
Could you please add an explicit LICENSE file to the repo so that it's clear under what terms the content is provided, and under what terms user contributions are licensed?

Per [GitHub docs on licensing](https://help.github.com/articles/open-source-licensing/#what-happens-if-i-dont-choose-a-license):

> Generally speaking, the absence of a license means that the default copyright laws apply. This means that you retain all rights to your source code and that nobody else may reproduce, distribute, or create derivative works from your work. This might not be what you intend.

Thanks!

davidchambers:
I believe the original author and all subsequent contributors will need to agree on a licence. This may take some time. ;)

qstrahl:
Paging @pangloss ... and while I'm at it, let's page @davidchambers and @amadeus and @goatslacker 

My usual go-to license for vim projects is to release it under the same terms as vim itself, as laid out under `:help license`. What do y'all think?

davidchambers:
> My usual go-to license for vim projects is to release it under the same terms as vim itself, as laid out under `:help license`.

Sounds good to me!

WChargin:
:+1: for my contributions. Let me know if I need to sign anything…

pangloss:
Oh, I thought there was a license in place... I'm happy with any open source license that you guys agree upon.

amadeus:
Fixed in `develop`

PR #339, opened 2015-12-31
Make arrow-functions one-liners

WChargin:
Summary:
This fixes the indentation for expression arrow functions
that are continued on multiple lines.
This does not affect the `FunctionBody` variant of arrow functions,
which were already correct due to the opening brace.

Resolves pangloss/vim-javascript#332.

Test Plan:
Type
```javascript
const myNiftyArrowFunction = (foo, bar) =>
    foo + bar;
myNiftyArrowFunction(1, 2);
```
and note that it's indented correctly as you type.
Then run `gg=G` and make sure that it stays indented correctly.

Note also that the indentation of
```javascript
const myFunctionBodiedArrowFunction = (foo, bar) => {
    doAThing();
    return foo + bar;
};
myFunctionBodiedArrowFunction(1, 2);
```
is still correct.

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -55,7 +55,7 @@ let s:continuation_regex = '\%([\\*+/.:]\|\%(<%\)\@<![=-]\|\W[|&?]\|||\|&&\|[^=]
 " TODO: this needs to deal with if ...: and so on
 let s:msl_regex = s:continuation_regex.'|'.s:expr_case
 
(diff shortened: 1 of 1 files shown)


WChargin:
In light of #338, I'm happy to release this code under the MIT license or any other (reasonable) choice.

davidchambers:
LGTM. Thanks!

PR #346, opened 2016-01-26
make the m variable script-local

artm:
otherwise it might conflict with somebody else's `m`


davidchambers:
Looks good, but please open a pull request against the __develop__ branch.

PR #348, opened 2016-01-29
Use shiftwidth() instead of &sw in indent/javascript.vim

zbinlin:
When sw set to 0, shiftwidth() will return tabstop value.

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -26,6 +26,17 @@ endif
 let s:cpo_save = &cpo
 set cpo&vim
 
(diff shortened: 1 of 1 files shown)


davidchambers:
I don't have enough context to review this change. Hopefully one of the other maintainers will chime in. :)

qstrahl:
This is fine; this is the recommended usage from `:help shiftwidth()`

I have to wonder why vim doesn't recommend you return `&sw ? &sw : &ts` if `shiftwidth()` doesn't exist, but this looks fine to me.

PR #353, opened 2016-02-23
fixed one line scope indent

camsteffen:
Continuing from #293. Fixed cases where the statement ends in parenthesis like `if(foo) bar()`


camsteffen:
I'm closing this because it breaks `if(foo())`

PR #354, opened 2016-03-05
Adding first pass at ES6 Class support

amadeus:
Been messing around with ES6 lately, so I decided to take the plunge and add proper Class support.

There is a good chance this may have bugs, or things that I have overlooked, however it appears to be working quite well so far.

This is before - as you can see (or not, if you aren't familiar with my color scheme), it's creating the method declarations within the class definition as actual function executions.
before

This is after - it properly interprets the function definitions and re-uses existing match groups. It also has the benefit of working in very bizarre contexts, as can be seen here (I probably haven't tested every possible scenario, so please let me know if you find issues)
after


amadeus:
There may also may be a good discussion surrounding what highlighting groups this should be matched with.

davidchambers:
Nice work! This is a huge improvement. :)

amadeus:
I've used this a bunch more, and it's appeared to work well. Any objections to merging this into Dev?

davidchambers:
No objections. :)

PR #358, opened 2016-04-04
Proper Lambda Support

amadeus:
This PR is a first pass at properly supporting ES6 lambdas. Currently we only support the `=>` (arrow) itself, and don't actually recognize the rest as a function declaration.

This first pass should handle the majority of use cases, although there may be some scenarios I missed.

For reference, this is what syntax highlighting looked like prior to this PR:

before

And this is how it looks with this PR applied: 

after

I also fixed a bug when defining a generator that would break `jsFuncArgs` highlighting

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -198,7 +198,7 @@ syntax region  jsBracket    matchgroup=jsBrackets     start="\[" end="\]" contai
 syntax region  jsParen      matchgroup=jsParens       start="("  end=")"  contains=@jsAll,jsOf,jsParensErrA,jsParensErrC,jsParen,jsBracket,jsBlock,@htmlPreproc fold
 syntax region  jsBlock      matchgroup=jsBraces       start="{"  end="}"  contains=@jsAll,jsParensErrA,jsParensErrB,jsParen,jsBracket,jsBlock,jsObjectKey,@htmlPreproc,jsClassDefinition fold
 syntax region  jsClassBlock matchgroup=jsClassBraces  start="{"  end="}"  contains=jsFuncName,jsClassMethodDefinitions contained fold
-syntax region  jsFuncBlock  matchgroup=jsFuncBraces   start="{"  end="}"  contains=@jsAll,jsParensErrA,jsParensErrB,jsParen,jsBracket,jsBlock,@htmlPreproc,jsClassDefinition contained fold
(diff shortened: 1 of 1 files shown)


syntax/javascript.vim
```diff
@@ -214,14 +214,18 @@ if main_syntax == "javascript"
...
 syntax match   jsFuncName       contained /\<[a-zA-Z_$][0-9a-zA-Z_$]*/ nextgroup=jsFuncArgs skipwhite
 syntax region  jsFuncArgs       contained matchgroup=jsFuncParens start='(' end=')' contains=jsFuncArgCommas,jsFuncArgRest,jsAssignmentExpr,jsComment,jsLineComment nextgroup=jsFuncBlock keepend skipwhite skipempty
 syntax match   jsFuncArgCommas  contained ','
 syntax match   jsFuncArgRest    contained /\%(\.\.\.[a-zA-Z_$][0-9a-zA-Z_$]*\))/
 
-exe 'syntax match jsArrowFunction /=>/ '.(exists('g:javascript_conceal_arrow_function') ? 'conceal cchar='.g:javascript_conceal_arrow_function : '')
+" Matches a single keyword argument with no parens
+syntax match   jsArrowFuncArgs  /\(\k\)\+\s\+\(=>\)\@=/ skipwhite contains=jsFuncArgs extend
+" Matches a series of arguments surrounded in parens
+syntax match   jsArrowFuncArgs  /(\(\k\|,\|\s\|\n\|\.\)*)\s\+\(=>\)\@=/ skipwhite contains=jsFuncArgs extend
```
davidchambers: Could you explain these two patterns?

amadeus: This is used to match the arguments definition for an `=>` function.

Currently for normal functions we actually highlight the parenthesis and actual arguments definitions, see this example:

func arg example

My color theme makes them orange.

To match this for lambdas, it's a bit trickier. We have to use a lookahead to match against the `=>`.

I separated it into 2 lines to make the overall regexes easier to understand.

Lambda functions can be both defined like this:

```js
// The first of the syntax match lines matches this case
arg => obj.getter();
```

And like this:

```js
// The second syntax match line matches for this
(arg) => obj.getter()
```

davidchambers: Thanks! What is `\k`?

amadeus: `\k` is for a keyword character, `:help \\k` for more deets

davidchambers:
LGTM

PR #359, opened 2016-04-04
Higher specificity for jsTemplate braces

amadeus:
I had the need for some more targeted syntax highlighting for template string braces, and having them namespaced to the default `jsBraces` group presented a problem.

This PR creates a new highlight group, `jsTemplateBraces`, that is then linked backed to `jsBraces` to maintain backwards compatibility.


davidchambers:
LGTM

PR #363, opened 2016-04-06
Misc Fixes

amadeus:
This PR is a few fixes we should get into `develop` before merging `develop` into `master`.

**Fixed Comment Regions**
* Fixed comment regions by adding the extend keyword (this means they override containing region endpoints, which they do when it comes to actual code. Before this PR, if you created a block comment inside say an object, with no closing block comment, the comment would appear to end at the end of the object. This is obviously not how block comments work.

oops

**Function Argument Declaration Destructuring Improvements**
* Some small improvements to destructuring in function declarations. This is far from complete, but a bit better than it used to be and handles many standard cases (I hope)

yay

**Fixed `jsBlock` vs `jsFuncBlock` priority in top level code**
* Before this PR, if you wrote an object at the top level of a file (i.e. not contained in a function or object or something), the braces would detect as `jsFuncBlock`. This bug was introduced from my earlier updates to arrow functions. The simple fix for this was to make `jsBlock` higher priority. `jsFuncBlock` continues to work normally in other contexts because specifiers like `nextgroup=` take priority.

oops2

### Note
We still have a major issue with arrow function declarations - allowing the argument parenthesis to span multiple lines. This is actually quite a challenging regex problem, and I don't think I will be able to come up with a fix in a reasonable amount of time and think we should ship the current `develop` branch with these fixes asap since there's a lot of great stuff in there.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -218,15 +218,16 @@ exe 'syntax match jsArrowFunction /=>/ skipwhite nextgroup=jsFuncBlock contains=
 
 syntax match   jsGenerator       contained '\*' nextgroup=jsFuncName,jsFuncArgs skipwhite
 syntax match   jsFuncName        contained /\<[a-zA-Z_$][0-9a-zA-Z_$]*/ nextgroup=jsFuncArgs skipwhite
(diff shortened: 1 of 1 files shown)


davidchambers:
> Fixed comment regions by adding the extend keyword (this means they override containing region endpoints, which they do when it comes to actual code.

This is wonderful! Thanks for the explanation.

amadeus:
Any qualms with merging this in? Otherwise I'll go ahead and do just that.

davidchambers:
No qualms. I won't claim to understand all the changes, though. ;)

amadeus:
Haha, no worries, it's weird to even explain some of them

PR #369, opened 2016-04-10
additional conceals: setTimeout, ===

szilveszter9:


amadeus:
I am not sure how I agree with adding all these extra conceal options. Personally I feel that all concealing stuff should perhaps be it's own file that could then have it's own batch of logic added to toggle it on or off.

Issue #329, opened 2015-12-10
Better ternary operator support


ksmithbaylor:
I use ternary expressions a lot, as part of a more functional "assign-once" style. But I always have to manually adjust my indentation and make sure to never accidentally auto-indent the parts of my code that use them. Here's the indentation I would like to use:

```javascript
// single-line values
const foo = condition ?
  trueValue :
  falseValue;
nextLineOfCode();
```
```javascript
// alternate style
const foo = condition
  ? trueValue
  : falseValue;
nextLineOfCode();
```
```javascript
// multi-line values
const foo = condition ? (
  moreComplex(
    multiLineStatement
  )
) : (
  otherMultilineStatement(
    ohNo
  )
);
```

However, the current implementation indents this code like so:

```javascript
// single-line values
const foo = condition ?
  trueValue :
    falseValue;
    nextLineOfCode();
```
```javascript
// alternate style
const foo = condition
  ? trueValue
  : falseValue;
  nextLineOfCode();
```
```javascript
// multi-line values
const foo = condition ? (
  moreComplex(
    multiLineStatement
  )
) : (
otherMultilineStatement(
  ohNo
)
);
```

iryan2:
This is something I regularly run in to as well, just thought I'd voice my support for this issue.

jslatts:
This is something that I run into as well. I have been running locally with this patch: jslatts/vim-javascript/b4b393646e54809cba5113f1a9d621ba564374ff

It gets close, but doesn't totally handle some of my use cases with JSX. I also don't know if it has any adverse side effects since I am not familiar with vim-script. Happy to make it a PR if the maintainers think it is ok.

robjens:
I guess this is the same indention problem I've tried to describe in #326

amadeus:
The first example case you have provided should be fixed in `develop`, the other ones are still issues.

Part of the challenge with this is that one size may not be able to fit all, to do it right there might have to be various modes that the indent file is put into, like comma first mode or whatever.

Unfortunately, none of the maintainers know much about the indent side of things, it was an inherited piece of work from previous maintainers that are no longer around, and thus is not updated often.

bounceme:
I'm for merging @jslatts solution as long as there aren't bad side effects

jslatts:
FWIW, I have been using my patch since February without any side effects (that I have noticed). It doesn't help most of the time because I tend to use the ternary statements with JSX elements which still format poorly.

bounceme:
[...]

PR #371, opened 2016-04-12
ternary indentation if out of expression issue

bounceme:


bounceme:
oh, i didn't see #329 and the available p.r. which is the same

Issue #364, opened 2016-04-06
jsParensError for parenthesized arrow function


davidchambers:
I just noticed that the closing parens in the following line is marked as `jsParensError`:

    f(() => {});
              ^

The same applies in this case:

    f(() => {
    });
     ^

This also exhibits the problem:

    f((x) => {});
               ^

Interestingly, this does not:

    f(x => {});

When I have a bit more time I will run `git bisect` to determine which change is responsible.

amadeus:
Good find, this is part of the challenge with arrow functions in general. Finding the specific commit is probably not necessary, I'll take a look.

davidchambers:
It was something in #363. I've played around a bit, but so far haven't determined exactly which change is responsible.

tomxtobin:
I just ran into this; it's 36f34b25d63211f1fab20c1f748a3dd070b4c270 according to `git-bisect`.

tomxtobin:
Reverting [this change](https://github.com/pangloss/vim-javascript/commit/36f34b25d63211f1fab20c1f748a3dd070b4c270?diff=unified#diff-af5b29be69c5b34c00e40f13b7e92435R230) fixes the issue:
```diff
-syntax match   jsArrowFuncArgs  /(\(\k\|,\|\s\|\n\|\.\)*)\s\+\(=>\)\@=/ skipwhite contains=jsFuncArgs extend
+syntax match   jsArrowFuncArgs  /(\%(.\)*)\s*\(=>\)\@=/ skipempty skipwhite contains=jsFuncArgs nextgroup=jsArrowFunction
```

tomxtobin:
It's the greedy `\%(.\)*` inside the arrow parameters' parentheses match. It's consuming too much of the line, including the closing `)` of the parameters.

A quick fix is changing that to the non-greedy equivalent `.{-}` — the Vim equivalent of PCRE's `.*?` — but then you're still matching unbalanced parentheses and whatnot. (We don't need the surrounding `\%(` and `\)` either way.)

davidchambers:
Thanks for investigating, @tomxtobin!

> A quick fix is changing that to the non-greedy equivalent `.{-}` — the Vim equivalent of PCRE's `.*?` — but then you're still matching unbalanced parentheses and whatnot.

[...]

PR #374, opened 2016-04-13
Fix arrow function parens arguments

dodev:
The syntax highlighter showed error when arrow functions had parentheses surrounding their arugemnts:
```javascript
...
  .then((result) => {
      send(result);
  })
--^ // The last parenthesis of `then` method call was considered excess.
``` 
I fixed the regex for args in parenthesis and it seems it works. I'm using it right now.


davidchambers:
This appears to fix #364. Please open a new pull request against the __develop__ branch, @dodev. Thanks!

PR #375, opened 2016-04-14
Fix arrow function parens arguments

dodev:
The syntax highlighter showed error when arrow functions had parentheses surrounding their arugemnts:
```javascript
...
  .then((result) => {
      send(result);
  })
--^ // The last parenthesis of `then` method call was considered excess.
``` 
I fixed the regex for args in parenthesis and it seems it works. I'm using it right now.

As seen in #374 :)


amadeus:
So, unfortunately, technically this doesn't work. It may appear to work based on your syntax highlighting, however it actually completely breaks the matching of `jsArrowFuncArgs`. In other words, it's the same as as just removing that entire match group completely.

I have been doing a lot of research into this, and it's actually a quite difficult problem to solve. On potential solution may be to remove the `jsArrowFuncArgs` highlight group completely until I can figure out a solution that works. I've tested a bunch of other JS syntax plugins, including even in other editors, and nobody can get it right for all scenarios, it's a real clusterfuck.

dodev:
TBH, I don't understand the regex syntax in viml and I thought it was a typo, a missing closing bracket. I hope somebody finds a solution. Unfortunately I don't have enough free time to look into this issue.

Issue #378, opened 2016-04-15
Function call brackets closes incorrectly when have ES6 arrow function as argument


unclechu:
I hope this screenshots will explain a lot:
![bug with ES6 arrow function](http://pix.toile-libre.org/upload/original/1460732419.png)
![works correctly with usual function](http://pix.toile-libre.org/upload/original/1460732448.png)
Here the text example:
- ES6 arrow bug: it**(**'_.blind(**)**', () => {}<em>)</em>;
- Usual function: it**(**'_.blind()', function () {}**)**;

amadeus:
Yup, known issue. We'll get a fix out shortly!

unclechu:
May be it's same issue:
![Issue demonstration screenshot](http://pix.toile-libre.org/upload/original/1460986854.png)
First `=>` wasn't highlighted.

amadeus:
Things should be sort of fixed for now in `develop`.

@unclechu I am working on another PR to bring some of the old functionality back, do you mind posting your screenshot snippets as code examples? So I can work them into a test file for the next update?

The more crazier examples you can provide that you actually end up using, would be super helpful!

jackfranklin:
@amadeus just wanted to say thanks for your work on this, and generally for your work on this repo. As a user it's awesome to see issues being fixed so quickly, and I really appreciate it :)

unclechu:
@amadeus 
```javascript
const f = func('_.foo()', () => {});
```
```javascript
const async = require('async');
const fs = require('fs');
async.waterfall([
  (cb) => cb(null, 'file.txt'),
  (path, cb) => fs.exists(path, (exists) => cb(null, exists)),
], (err, exists) => if (err) console.error(err); else console.log('exists:', exists));
```

unclechu:
@amadeus 
```javascript
const found = _.find(headers, (it) => /^Content-Length: \d\+$/i.test(it));
```

amadeus:
@unclechu these are great, thanks!

amadeus:
This issue should be fixed in `develop`

unclechu:
Also arrow function single argument without brackets example:
```javascript
const found = _.find(headers, it => /^Content-Length: \d\+$/i.test(it));
```
```javascript
const async = require('async');
const fs = require('fs');
async.waterfall([
  cb => cb(null, 'file.txt'),
  (path, cb) => fs.exists(path, (exists) => cb(null, exists)),
[...]

PR #380, opened 2016-04-18
Temporarily disabled jsArrowFuncArgs with parens

amadeus:
This is until we can figure out how to properly match all the crazy use
cases.

This should at least make issues #364 and #378 not appear as errors.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -227,7 +227,7 @@ syntax match   jsFuncArgDestructuring contained /\({\|}\|=\|:\|(\|)\|\[\|\]\)/ e
 " Matches a single keyword argument with no parens
 syntax match   jsArrowFuncArgs  /\(\k\)\+\s*\(=>\)\@=/ skipwhite contains=jsFuncArgs nextgroup=jsArrowFunction
 " Matches a series of arguments surrounded in parens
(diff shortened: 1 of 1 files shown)


syntax/javascript.vim
```diff
@@ -227,7 +227,7 @@ syntax match   jsFuncArgDestructuring contained /\({\|}\|=\|:\|(\|)\)/ extend
 " Matches a single keyword argument with no parens
 syntax match   jsArrowFuncArgs  /\(\k\)\+\s*\(=>\)\@=/ skipwhite contains=jsFuncArgs nextgroup=jsArrowFunction
 " Matches a series of arguments surrounded in parens
-syntax match   jsArrowFuncArgs  /(\%(.\)*)\s*\(=>\)\@=/ skipempty skipwhite contains=jsFuncArgs nextgroup=jsArrowFunction
+" syntax match   jsArrowFuncArgs  /(\%(.\)*)\s*\(=>\)\@=/ skipempty skipwhite contains=jsFuncArgs nextgroup=jsArrowFunction
```
davidchambers: We reference `jsArrowFuncArgs` in other places. Will this cause problems since it is no longer defined?

amadeus: There's a `jsArrowFuncArgs` right above it, the single keyword detection is very safe and I am leaving that in, so ti shouldn't matter. I am keeping this as a comment for easy reference for to fix.

davidchambers:
LGTM

PR #381, opened 2016-04-19
Better handling of multiline ternary indents. #329

jslatts:
As requested in #329, here is a PR with my change.

One thing I noticed when testing again, is that the indenting does not working correctly while typing out the multi-line ternary statement, but on re-identing the block, it does. However, this behavior is the same without my changes as well.

I read through the code again and I don't see any obvious reason why it would behave this way. If someone more familiar with the code can point me in the right direction, I'm happy to investigate further.

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -78,6 +78,7 @@ let s:comma_last = ',\s*$'
 
 let s:ternary = '^\s\+[?|:]'
 let s:ternary_q = '^\s\+?'
+let s:ternary_s = '^\s\+:'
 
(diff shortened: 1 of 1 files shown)


amadeus:
I'm good with getting this in, anyone object?

bounceme:
Sorry, my new request already fixes this

amadeus:
So you're saying we should merge in #383 and not this one?

bounceme:
Yea, it can be merged soon, i'll let you know when it's done.There are a few bugs still

amadeus:
Closing due to #385

PR #379, opened 2016-04-16
simplifies the parentheses handling

bounceme:
also a possible fix for the last example in #329 and #315.
this still needs some work

PR #383, opened 2016-04-19
Operator first support,including ternary,dot etc.

bounceme:
generally this will give us fairly accurate indents without having specific code for each operator.Thanks to https://github.com/pangloss/vim-javascript/issues/139#issuecomment-55064913 which this is based on
I think this needs a few things before it's ready


bounceme:
Oh, this needs #379

bounceme:
This should be finished!

PR #384, opened 2016-04-19
Fixing comments in ES6 Class declarations

amadeus:
This is a pretty big durp and a painfully stupid fix


amadeus:
Since this is a pretty big bug, and a painfully easy fix, I'm going to just go ahead and get this into `develop` ASAP.

PR #388, opened 2016-04-20
Improved Lambda Argument Handling

amadeus:
This is a first pass at adding back in the `jsArrowFuncArgs` matching when using parenthesis.

It should mostly fix issues specified at #378 and #364.

I have been using the following test file to ensure syntax highlighting seems to work, I would be open to more test cases to add to this to ensure things are working.

```js
/* eslint vars-on-top:0 */
function test1({size = 'big', cords = { x: 0, y: 0 }, radius = 25} = {}) {
	console.log(size, cords, radius);
}

var test2 = function(test2arg = () => {
	return true;
}) {

};

var test3 = function(){};

var test4 = (test4arg, { test4arg2 = 2 }) => {};

test3((arg) => {
	var test5 = (test5arg) => {
	};
});

test3((test6arg = 2 + 2, { test6arg2: x, test6arg3: y }) => {
});

test3((test7arg) => {});

test3(test8arg => {});

var test9 = function(x, y){
	var tub = () => {};
};

var test10 = (test10arg, /* test comment */ test10arg2 = () => {}) => {};

var test11 = {};

test3.then(
	(result) => {
		var test12 = (test12arg = { test12arg2: [] }) => {};
		this.send(result);
	}
);

test3('_.blind()', () => {});

test3((test13arg, test13arg2) => test4.exists(test13arg, (test14arg) => {
	return true;
}));

// Checking syntax highlighting on the line before this, should be empty
```

We don't handle this file perfectly, but we do breakdown 'gracefully' (meaning we do not show error parenthesis groups).

Current issues I haven't been able to figure out:

* How to handle parenthesis inside the `jsArrowFuncArgs` match
* Handle a multiline version of `jsArrowFuncArgs`

I have done a bit of research into other editors like Sublime and Atom, and none can even handle this file completely correctly, and they ALL break with the 2 issues I mentioned above (at least their default syntax highlighting, I never explored plugins that may do things better).


davidchambers:
Nice work, @amadeus! I'm pleased to see how minimal the changes are.

amadeus:
I've gone ahead and push a copy of this branch to this repo, so it's easier for you guys to test:

https://github.com/pangloss/vim-javascript/tree/lambda-experiments

alexlafroscia:
@amadeus I tried out that branch, it fixed some issues I was seeing around highlighting getting messed up when using anonymous "fat arrow" functions, where the parameters are surrounded in parenthesis.  Thanks!

Issue #324, opened 2015-11-24
Regex ending in * breaks indenting in block below it


eddiezane:
It appears a regex ending in `*` breaks indenting below it in a block.

```javascript
if (true) {
  var re = /.*/;
console.log('I am broken');
console.log('Because I end in *');
}

if (true) {
  var re = /.*$/;
  console.log('I do not end in *');
  console.log('So I am fine');
}
```

Can anyone replicate or confirm?

Apologies if this was already reported.

davidchambers:
I can't reproduce the problem. Could you disable your plugins one by one to confirm vim-javascript is responsible?

eddiezane:
Reproduced it with a completely blank vimrc aside from vim-javascript.

Try running a `gg=G` on it in a buffer?

davidchambers:
> Try running a `gg=G` on it in a buffer?

Ah, now I see it! Thanks for taking the time to verify that vim-javascript is at fault.

The problem is that none of this project's current maintainers is familiar with the indentation-related code.

amadeus:
@bounceme it appears some of the fixes we have in place, do not function when `gg=G` is used. Everything works fine when you type out the examples manually however. Any ideas?

amadeus:
This issue should be fixed in `develop`, by #390

PR #390, opened 2016-04-21
#324

bounceme:
This should work,some things still aren't correct if you abuse multiline comments like https://github.com/Khan/tota11y/blob/master/build/tota11y.js


amadeus:
Looks good to me, thanks!

PR #392, opened 2016-04-22
Fix jsDocParam

amadeus:
I am not sure exactly why this problem exists, but it causes some weird side effects.

Before this commit:
before

After this commit:
after

The arrows point out the syntax errors, and there's a deeper problem. If you use `%` on the opening `{` of the function, it will match with the second `}` in the comment. It should of course match the final closing `}` of the function.

The fix I have for it, is to only match on paired opening and closing curly brackets. I have not noticed any issues to existing JSDoc comments.


amadeus:
And here's a copy/pastable snippet from the screenshot if you want to test yourselves:

```js
(function(){

var testOb = {

    /**
    * test text {@link #add} or {@link #addOnce}.
    * @property {boolean} memorize
    */
    memorize: false,
};

})();
```

amadeus:
I haven't noticed any side effects from this pr, so going ahead and merging it in since it's pretty basic.

PR #400, opened 2016-04-25
Switch mechanics, indent

bounceme:
also now that I know about cindent i'm curious if there are other places like this to remove code and increase performance

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -65,10 +65,6 @@ let s:line_term = '\s*\%(\%(\/\/\).*\)\=$'
 " Regex that defines continuation lines, not including (, {, or [.
 let s:continuation_regex = '\%([\\*/.:]\|+\@<!+\|-\@<!-\|\%(<%\)\@<!=\|\W[|&?]\|||\|&&\|[^=]=[^=>].*,\)' . s:line_term
 
-" Regex that defines continuation lines.
-" TODO: this needs to deal with if ...: and so on
-let s:msl_regex = s:continuation_regex.'\|'.s:expr_case
-
 let s:one_line_scope_regex = '\%(\%(\<else\>\|\<\%(if\|for\|while\)\>\s*(\%([^()]*\|[^()]*(\%([^()]*\|[^()]*(\%([^()]*\|[^()]*([^()]*)[^()]*\))[^()]*\))[^()]*\))\)\|=>\)' . s:line_term
 
 " Regex that defines blocks.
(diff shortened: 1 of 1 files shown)


amadeus:
Testing this out, I am noticing some weirdness

If I just use the indentation given to me, I get this scenario:

```js
switch (test) {
	case 1:
	test = 2;
	word = 3;
	hi = 4;
	break;
	case 2:
	weird = 2;
	hi = 3;
	break;
	default:
	test = 10;
	word = 20;
	break;
}
```

Yet if I try to indent the first line, 2 different things  will happen:

```js
switch (test) {
	case 1:
		test = 2; // purposely indented this line
	word = 3;
	hi = 20;
	break;
	case 2:
		hi = 2; // Purposely indented this line
		weird = 2;
		strange = 3;
		break;
	default:
		weird = 3; // Purposely indented this line
		hi = 2;
		break;
}
```

bounceme:
Your example is actually broken for me too. If I understand right, this is being caused by something to do with the breaks

bounceme:
This seems to work now and has removed even more code

indent/javascript.vim
```diff
@@ -475,11 +449,7 @@ function GetJavascriptIndent()
 
   " If the previous line ended with a block opening, add a level of indent.
   if s:Match(lnum, s:block_regex)
-    if (line =~ s:expr_case)
-      return indent(s:GetMSL(lnum, 0)) + s:sw()/2
-    else
       return indent(s:GetMSL(lnum, 0)) + s:sw()
```
davidchambers: :arrow_left: :arrow_left:

davidchambers:
I :heart: all the red in the diff!

bounceme:
also people could customize this in a few ways:
```
set cino+=:0
```

amadeus:
Unfortunately this appears to have many more issues.

First off, here's the default indents if I don't change anything:

```js
switch (case) {
    case 1:
        test = 1;
    another = 2;
    hi = 3;
case 2:
    anotherTest = 1;
hmmm = 3;
break;
case 4:
    test = 1;
weird = 3;
}
```

It gets worse if I try to fix the indent:

```js
switch (case) {
    case 1:
        test = 1;
        test2 = 2; // this line had to be fixed
        test = 3;
    break; // this line had to be un-indented
case 2: // this line autoindents BACK when I type the :
}
```

amadeus:
Oh wait, hold on, I had the wrong PR, one sec

amadeus:
Would you like to hold off on merging this to find some other improvements you can make? Or would prefer to file those as separate PRs?

bounceme:
i'd prefer fixing whatever bugs come of this as they are found, which with my testing of this, isn't so worrying to prevent a merge.

PR #403, opened 2016-04-27
Custom jsFuncName for object definitions

### LINKED ISSUE #261: Support for ES6 class / object functions
ES6 has another syntax for functions. What was `{ someProp: function () {} }` can now be written as `{ someProp () {} }`.

screen shot 2015-04-03 at 6 18 09 pm

They should be `jsObjectKey`, but currently they get parsed as `jsFuncCall`. I tried my hand at adding support but couldn't grok vim's syntax highlighting.

amadeus:
This ES6 version of `jsFuncName` requires some special handling to not, by default, overwrite `jsObjectKey` and also not overwrite `jsFuncCall`.

This should fix #261 

I also fixed a bug in generator syntax matching within ES6 classes.


amadeus:
Still need to figure out some issues when using a generators asterisk, so plz do not merge yet

PR #404, opened 2016-04-27
misc fixes

bounceme:
I'll test this some more and get some examples showing the new parts

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -360,7 +360,7 @@ function GetJavascriptIndent()
       if line[col-1]==')' && col('.') != col('$') - 1
         let ind = virtcol('.')-1
       else
-        let ind = indent(s:GetMSL(line('.'), 0))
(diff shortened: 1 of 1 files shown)


bounceme:
```
one()
  .then(two)
  .then(function() {
    console.log("value");
  })
foo(a(b,
      c))

var protocol = self.proxy && !self.tunnel ? self.proxy.protocol : self.uri.protocol,
  defaultModules = {
    'http:': http,
    'https:': https
  },
  httpModules = self.httpModules || {}

self.httpModule = httpModules[protocol] || defaultModules[protocol]
```
This mostly just fixes a few odd kinds of behavior with the logic of chaining operators and commas. I've done a good amount of testing and saw no problems with perf or regressions

davidchambers:
> This mostly just fixes a few odd kinds of behavior with the logic of chaining operators and commas.

Sounds good!

bounceme:
I think this is merge-able. we can always revert if we need too

PR #408, opened 2016-04-29
multiline comment

bounceme:
this could be a fix


bounceme:
I could merge this now unless there are any concerns,nothing unusual caused by these changes from what I see

Issue #411, opened 2016-05-01
indentation bug


budkin:
If there is a parenthesis between quotes within an `if` or `while` statement the next line is not indented, as in `if (c = "(")`
https://i.imgur.com/PEzpSwW.png

budkin:
also, if a `case` in a `switch` statement contains an `if` or `while` the next line is unindented like it expects another `case`
https://i.imgur.com/1jIqKtt.png

bounceme:
The second issue should be corrected in develop, can you confirm this?i will be try  to replace the regex based parantheses detection

budkin:
Yes, the second issue is resolved in develop branch, switched to it, thank you

bounceme:
On the parentheses issue, a solution to this would require several changes throughout the indent script,I invite anyone to enlist and do it otherwise you may be waiting a while, I recommend reading ```h search pair```

PR #412, opened 2016-05-02
parentheses regex is thankfully gone

bounceme:
solves #411


bounceme:
with nothing looking wrong in my testing I'll merge to develop tomorrow

budkin:
thank you for you work

PR #414, opened 2016-05-02
Temporary fix for older versions of vim

amadeus:
Will need to better fix this soon, just getting this hotfix out asap


bounceme:
why aren't all the regexes in this file surrounded with quotes,is there any difference between them?

PR #417, opened 2016-05-03
Prevent limited lookbehinds in older vers of vim

amadeus:
I went the route of just `if` casing 2 different versions of the `jsRegexpString` regex based on the version of Vim you are running. I am generally not a fan of executing string concatenated regexs, but I am opening to this as a possibility if ya'll disagree.


bounceme:
since it's just one line it is fine with me,thanks!

Issue #86, opened 2013-07-12
Indentation shift


aymericbeaumet:
By now, the indentation is made this way:

```javascript
for (var i in test)
  {
    alert('test');
  }
```

I would like to do it like so:

```javascript
for (var i in test)
{
  alert('test');
}
```

I don't find any customization variables in the plugin. How can I do?

PR #426, opened 2016-05-08
c style opening curly braces

bounceme:
#86
One way to fix this, assuming this is after 'function()' or 'if()', are there other places you might see just a opening brace by itself?


bounceme:
```
for (var i in test)
{
  alert('test');
}

var i =
  {
    j:1
  }
```
there might be a few issues still

PR #436, opened 2016-05-18
Forcing `class` text to be a keyword

amadeus:
This class regex will match if the text class is contained within a larger variable name, and therefore is broken.

I.E. In a definition for `className`, `class` will match as `jsClassDefinition` which is bad.

This P.R. fixes that by forcing `class` to be a keyword.


amadeus:
This is a pretty bad bug and the fix is pretty simplistic. Any concerns against merging this? Would like to get it in, ASAP.

bounceme:
I am fine with it, maybe specify whitespace instead of word boundaries though, i don't really know the class syntax though

amadeus:
Unfortunately that isn't viable, since `class{}` is a valid class definition and don't want to break that.

bounceme:
yeah, word boundary is enough then

PR #440, opened 2016-05-23
wrapping comments improved

bounceme:
this improves most regexes and the prevnonblanknonstring function


bounceme:
i'm guessing this is finished

PR #442, opened 2016-05-25
Develop -> Master

amadeus:
Just another base release. I kinda want to get this out soon though, so we stop getting those bug reports regarding the lookbehind limiters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# vim-javascript v0.10.0
+# vim-javascript v1.0.0
 
 JavaScript bundle for vim, this bundle provides syntax highlighting and
 improved indentation.
--- a/README.md
+++ b/README.md
@@ -1,24 +1,8 @@
 # vim-javascript v0.10.0
 
-JavaScript bundle for vim, this bundle provides syntax and indent plugins.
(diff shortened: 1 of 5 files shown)


davidchambers:
Let's do it. :)

bounceme:
the indent script is ready for a merge

amadeus:
i added a small issue I noticed in #443, is that an easy fix? If so, I'd like to get that in first, otherwise we can hold off and fix later.

bounceme:
hmm, I'm at work. I'll try and fix it soon

bounceme:
actually, since we use cindent for multiline comments, this is probably customizable. maybe in the cinoptions section of the vim help you can find it

bounceme:
```
cN    Indent comment lines after the comment opener, when there is no
	      other text with which to align, N characters from the comment
	      opener.  (default 3).  See also |format-comments|.

		cino=			  cino=c5
		  /*			    /*
		     text.			 text.
		   */			     */
 ```

amadeus:
Ok cool, then I think we can just go ahead and merge this, and I'll worry about that later.

amadeus:
Alright, I think this may be good to go?

PR #443, opened 2016-05-25
another edge side effect

bounceme:
this is just a more correct, thorough check, for multiline comments


bounceme:
I'll merge if people are alright with this, it should go into the master merge

amadeus:
One weird thing I am noticing with block comments, if I delete the special additional asterisk, and then just do a normal tab, that line is fine, but when I press `<cr>`, I get a weird 3 space indent...

amadeus:
example

Here's a screenshot example

PR #445, opened 2016-05-26
a comma should be considered an operator

bounceme:
this fixes some serious issues with the way a continuation is detected and also removes the specific code for comma first style,which was very breakable


bounceme:
this should be worked on, it's not ready yet

bounceme:
I think i am satisfied with this.we have to get this into develop before the master merge.

PR #446, opened 2016-05-26
Shebang doesn't have to contain env

dNItro:
Skimming our neighbor's ( [vim-javascript-syntax](https://github.com/jelera/vim-javascript-syntax) ) commits saw this one. added `\%(\)` group catching to perform a little bit faster.


amadeus:
Can you please open commits against `develop`? That will also mean basing the commit off `develop`.

PR #447, opened 2016-05-26
regex warning

bounceme:


bounceme:
https://groups.google.com/forum/#!searchin/vim_dev/7.4.002/vim_dev/GfSwrDEAejM/zqgZJRHo1AEJ
this is the patch which fixes the issue with highlighting right? Also regexpengine is a global option

ftplugin/javascript.vim
```diff
@@ -1 +1,4 @@
 setlocal suffixesadd+=.js
+if v:version > 704 || (v:version == 704 && has('patch002'))
```
amadeus: Wait, this says if versions is GREATER than 704, shouldn't it LESS THAN?

ftplugin/javascript.vim
```diff
@@ -1,4 +1,4 @@
 setlocal suffixesadd+=.js
-if v:version > 704 || (v:version == 704 && has('patch002'))
+if (v:version < 704 || (v:version == 704 && has('patch002'))) && exists('&regexpengine')
```
amadeus: We need a NOT on the `has()` patch, since patch fixes, and therefore, we don't want `re=1` on `has()` being truthy.

bounceme:
how about now?

amadeus:
Looks good to me. Can you squash these commits locally though, and give them a better name before merging?

bounceme:
sorry,the pr name should be changed? you can squash and edit/remove the commit messages when merging on github now

PR #453, opened 2016-05-31
timeout the paren searching

bounceme:
the search pair function almost stalls my decently powerful macbook pro on really long lines, could be annoying because of minification etc.this function just makes it backwards compatible in old vim versions or builds without this feature.it is set to 300ms, though i am open to suggestions

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -59,7 +59,7 @@ let s:syng_linecom = 'linecomment\c'
 " Expression used to check whether we should skip a match with searchpair().
 let s:skip_expr = "synIDattr(synID(line('.'),col('.'),1),'name') =~ '".s:syng_strcom."'"
 
-func s:searchpar(start,mid,end,flags,stop)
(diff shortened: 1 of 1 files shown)


indent/javascript.vim
```diff
@@ -59,6 +59,14 @@ let s:syng_linecom = 'linecomment\c'
 " Expression used to check whether we should skip a match with searchpair().
 let s:skip_expr = "synIDattr(synID(line('.'),col('.'),1),'name') =~ '".s:syng_strcom."'"
 
+func s:searchpar(start,mid,end,flags,stop)
```
davidchambers: This name is confusingly similar to `searchpair`. Perhaps `s:searchparens` would be clearer.

davidchambers:
I look forward to this change. :thumbsup:

bounceme:
yes, I agree on the name

bounceme:
haha, just tried a skip_expr which instead match the paren syntax items instead of Not being a comment/string/regex. sadly it's just as slow

PR #455, opened 2016-06-01
limiting for slow function and operator fix

bounceme:
```
var magicWords = ['abracadabra', 'gesundheit', 'ventrilo'],
  spells = {
    'fireball': function() {
      setOnFire()
    },
    'water': function() {
      putOut()
    }
  },
  a = 1,
  b = 'abc',
  etc, somethingElse
someFunc(param1, param2,
  param3, param4)
```
at the line which starts with a curly brace followed by a comma this was needed for traversing the comma less property lines, however this only needs to be called once.this improves performance when using the endline operator style

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -383,7 +383,8 @@ function GetJavascriptIndent()
       if parlnum > 0 && s:Match(parlnum, s:operator_first)
         return indent(parlnum)
       end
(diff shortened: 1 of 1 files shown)


bounceme:
this is one of the really confusing parts about the indenter

bounceme:
i think this is good enough for a merge to dev

PR #457, opened 2016-06-01
refactoring

bounceme:
by making the comma behave as an operator we can remove all of the assumptive variable chain code,along with huge performance gains. experimental right now


bounceme:
i'm probably going to give up on the refactoring for now

PR #463, opened 2016-06-04
changes according to #466

bounceme:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors#Dot_notation

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -423,7 +423,7 @@ function GetJavascriptIndent()
 
   " If the previous line ended with a block opening, add a level of indent.
   if s:Match(lnum, s:block_regex)
-    return s:InMultiVarStatement(lnum, 0, 0) ? indent(lnum) + s:sw() : indent(s:GetMSL(lnum, 0)) + s:sw()
+    return indent(lnum) + s:sw()
   endif
 
   " Set up variables for current line.
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -384,7 +384,7 @@ function GetJavascriptIndent()
         return !s:Match(parlnum, s:operator_first) ? indent(lnum) + s:sw() : indent(parlnum)
       end
     elseif line !~ s:line_pre .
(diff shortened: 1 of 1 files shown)


bounceme:
and an empty string is a valid property name? hah

bounceme:
possibly adding basic computed properties also

bounceme:
@amadeus is there a way to get a specific syntax group like jsblock but for objects only?

amadeus:
@bounceme soooo, that's an update I am planning to do, and we will need it, especially for many ES6 features. Perhaps I can find some time tomorrow to look into it.

bounceme:
thanks, all the new shorthand syntaxes are dependent on that, otherwise my giant regex matches basically every object literal property definition in es5

bounceme:
i'll finish this with the es6/7 pr in mind. @amadeus i don't think the jsfuncargs group works across lines, is that something you would be able to fix?

amadeus:
`jsFuncArgs` should work just fine over multiple lines. For example:

```js
var func = function(
    x,
    y
){
    return x + y;
};
```

If you are referring however to `jsArrowFuncArgs`, then yes, unfortunately this is a limitation I don't think we can get around.

The TL;DR is that `jsArrowFuncArgs` can only be matched by a test on the final `)` to detect a `=>`. The way vim handles `syntax regions`, it will just start the region regardless if it can find an ending match which means it would conflict with our `jsParens` region.

No other syntax file that I have found actually supports a multiline `jsArrowFuncArgs` like match, they all break the moment you create a new line within them.

bounceme:
oh sorry, I mean is there an 'invoked function arguments' syntax pattern?

amadeus:
@bounceme ahh, at this point no, although we could potentially chain something to the `jsFuncCall` logic if that would help? It could have some weird side effects tho.

bounceme:
it isn't necessary really, jsparen is enough for now at least

bounceme:
this pr was waiting on #466, a relatively minor update but the additions are helpful for objects handling

bounceme:
going to merge this as it goes along with @amadeus 's update

PR #464, opened 2016-06-05
character classes are faster

bounceme:


bounceme:
this may be a minuscule improvement, syntax contributors may also prefer the collections in which case i can close this

amadeus:
I am not sure how far I will get, but I am looking into refactoring the syntax file for all these new ES6 aspects. It's gonna be quite a big change, and so these would be hard to rebase into those changes. Let me see how far I get today and if I get something workable, perhaps we look manually merging these changes in then.

bounceme:
if we get around to this, the regexes can be even cleaner using the `\k \K` classes. this brief form would work since we add $ to iskeyword

Issue #433, opened 2016-05-12
Syntax support for ES property initializers


iansinnott:
I'm not sure if this is a bug or simply something that is not yet supported, but it would be awesome to see support for [ES class property initializers](https://github.com/jeffmo/es-class-fields-and-static-properties). They are currently a stage-1 proposal but they make authoring React components much nicer.

I've tried this on both `develop` and `master`.

Example code:

```js
export class About extends React.Component {
  static propTypes = {
    children: PropTypes.any,
  }
  render() {
    return (
      <div className={cx('page')}>
        <div className={cx('siteTitle')}>
          <h1>About Page</h1>
        </div>
        <p>Welcome to the about page...</p>
      </div>
    );
  }
}

export class NotFound extends React.Component {
  render() {
    return (
      <div className={cx('page')}>
        <h4>Not found</h4>
      </div>
    );
  }
}
```

Here's how it's currently being highlighted. I annotated the screenshot with the syntax IDs of the highlighted parts of text:

![app_js______dev_js_redux-second-run_client_components__-_vim](https://cloud.githubusercontent.com/assets/3154865/15224383/e3e3ab64-182d-11e6-9249-caeddcb5da84.png)

Related: #393

amadeus:
Should be fixed in `develop`

Issue #425, opened 2016-05-08
Incorrect highlighting when defining a computed method


paxunix:
```
class Foo {
    [someThing]() {
        return 10;
    }
}
```
The current highlighter thinks the method body is jsClassBlock so its entire highlighting is thrown off.

amadeus:
I believe this is some form of ES7 level functionality? I will look into it as I look to add ES7 support. Thanks for posting the issue!

paxunix:
Computed method names (computed properties) is an ES6 feature:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

amadeus:
In objects yes, I believe you are correct. Curly brackets in class definitions have a different set of rules that are being expanded upon in ES7

amadeus:
Should be fixed in `develop`

Issue #393, opened 2016-04-22
Object on ES7 Class Highlights Closing Bracket Red


daGrevis:
![Preview][preview]

[preview]: http://i.imgur.com/h0T0Oix.png

When you define an object on ES6 class, closing bracket of that object will get highlighted in red. When you define a number, a string, an array or a lambda instead, everything works.

```javascript
class Thing {
    foo = {}
}

class OtherThing {
    bar = 42
}
```

You can reproduce this on `master` branch.

eddiezane:
Also reproducible on `develop`

amadeus:
I believe this is actually ES7 functionality? Which we've slowly been adding, but we are a bit behind there.

xwendotnet:
I have same problem when lambda as second parameter.

 
![screen shot 2016-04-23 at 10 16 59](https://cloud.githubusercontent.com/assets/7343175/14758748/97994912-093c-11e6-8bea-940779842566.png)

eddiezane:
@xwendotnet think your issue is different and fixed `develop`.

amadeus:
@xwendotnet yes, that is a different issue and what @eddiezane said, it should now be fixed!

amadeus:
*in the 'develop' branch

xwendotnet:
@amadeus thank you, I will checkout that branch

daGrevis:
> I believe this is actually ES7 functionality?

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes): “JavaScript classes are introduced in ECMAScript 6“.

> Which we've slowly been adding, but we are a bit behind there.

Thanks!

amadeus:
@daGrevis I should have been more clear. We support ES6 classes. The syntax here is actually new stuff added into ES7. If you look through that entire article, you won't find any examples of 'prop = val'

ES6 Classes only support functions in the definition.

daGrevis:
Ah, I see. My bad. :smile:

amadeus:
Should be fixed on `develop`

PR #466, opened 2016-06-06
ES6/7 Refactor

amadeus:
Ok, this one has been a long time coming, and I finally found some time to sit down and work on it.

This PR refactors quite a bit of the syntax file to allow us to better support both current ES6 features, and make it far easier to support upcoming ES7 features.

The big thing it does is actually create many more distinctions around block types. By block types I mean anytime you use `{` and `}`.  This includes being able to detect an if statement, vs a switch statement, vs a for/do/while loop, vs an actual object definition, vs function blocks, etc.

There are probably going to be some bugs.

One thing this PR does do, is provide a fix/support for:

```js
var obj = {
    func(){},
    func2(){},
    [CONST.SOMETHING],
    [CONST.SOMETINGELSE]: function(){}
};
```

It also lays the framework for me to fix/add support for #433, #425, #393

I've made this branch on this repo, so please check it out and try using it so we can get it a bit more battle tested before merging it in.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -93,7 +93,7 @@ exe 'syntax keyword jsThis      this   contained '.(exists('g:javascript_conceal
 exe 'syntax keyword jsSuper     super  contained '.(exists('g:javascript_conceal_super')     ? 'conceal cchar='.g:javascript_conceal_super      : '')
 
 " Statement Keywords
-syntax keyword jsStatement    contained break continue with yield
+syntax keyword jsStatement    contained break continue with yield debugger
 syntax keyword jsConditional            if else      skipwhite skipempty nextgroup=jsParenIfElse,jsBlock
 syntax keyword jsConditional            switch       skipwhite skipempty nextgroup=jsParenSwitch
 syntax keyword jsRepeat                 while for    skipwhite skipempty nextgroup=jsParenRepeat
@@ -109,7 +109,7 @@ syntax keyword jsGlobalObjects  Array Boolean Date Function Iterator Number Obje
 syntax keyword jsExceptions     Error EvalError InternalError RangeError ReferenceError StopIteration SyntaxError TypeError URIError
 syntax keyword jsBuiltins       decodeURI decodeURIComponent encodeURI encodeURIComponent eval isFinite isNaN parseFloat parseInt uneval
 " DISCUSS: How imporant is this, really? Perhaps it should be linked to an error because I assume the keywords are reserved?
-syntax keyword jsFutureKeys     abstract enum int short boolean interface byte long char final native synchronized float package throws goto private transient debugger implements protected volatile double public
+syntax keyword jsFutureKeys     abstract enum int short boolean interface byte long char final native synchronized float package throws goto private transient implements protected volatile double public
 
 " DISCUSS: Should we really be matching stuff like this?
 " DOM2 Objects
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -32,7 +32,7 @@ syntax match   jsParensError    /\%()\|}\|\]\)/
 syntax keyword jsStorageClass   const var let
 syntax keyword jsOperator       delete instanceof typeof void new in of
 syntax match   jsOperator       /[\!\|\&\+\-\<\>\=\%\/\*\~\^]\{1}/
-syntax match   jsSpreadOperator /\.\.\./
+syntax match   jsSpreadOperator /\.\.\./ skipwhite skipempty nextgroup=@jsExpression
 syntax keyword jsBooleanTrue    true
 syntax keyword jsBooleanFalse   false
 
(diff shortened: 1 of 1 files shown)


amadeus:
I've also reworked the file a bit to make it more organized, and tried to add more consistency with how we organize arguments and stuff. Please do not merge anything related to syntax files in until this gets merged in, or else it might be a nightmare for me to fix the merge conflicts.

bounceme:
hey, just tried it out. by a quick test with indenting,since it relies on some syntax items, there seems to be a difference with line-comments. I couldn't pretend to be a great judge on the syntax side of things but i think you have done a great job! also seems faster, even with indenting

amadeus:
I actually removed line comments as an item and just made them part of `jsComment`. Is this going to mess things up for you? I could put that back in.

bounceme:
i'll try a few things out in terms of the line comment detection, i don't know for sure but maybe line comments are simple enough for a regex in the indent script

amadeus:
I've gone ahead and made a first pass at supporting/fixing the following issues: #433, #425, #393

@iansinnott, @daGrevis, @paxunix if you guys can check out the `es6-refactor` branch and see how it fixes your issues, that would be super helpful. There will probably be certain bugs or things I have not yet forseen that still need to be fixed, but the more eyes on this the better.

amadeus:
@bounceme @davidchambers I added a couple `DISCUSS` comments in the changes, would be curious to get your thoughts on it as well.

syntax/javascript.vim
```diff
@@ -16,213 +16,193 @@ if !exists('g:javascript_conceal')
...
+syntax keyword jsAsyncKeyword           async await
+
+" Keywords
+syntax keyword jsGlobalObjects  Array Boolean Date Function Iterator Number Object Symbol Map WeakMap Set RegExp String Proxy Promise Buffer ParallelArray ArrayBuffer DataView Float32Array Float64Array Int16Array Int32Array Int8Array Uint16Array Uint32Array Uint8Array Uint8ClampedArray JSON Math console document window Intl Collator DateTimeFormat NumberFormat
 syntax keyword jsExceptions     Error EvalError InternalError RangeError ReferenceError StopIteration SyntaxError TypeError URIError
-
 syntax keyword jsBuiltins       decodeURI decodeURIComponent encodeURI encodeURIComponent eval isFinite isNaN parseFloat parseInt uneval
-
+" DISCUSS: How imporant is this, really? Perhaps it should be linked to an error because I assume the keywords are reserved?
 syntax keyword jsFutureKeys     abstract enum int short boolean interface byte long char final native synchronized float package throws goto private transient debugger implements protected volatile double public
```
davidchambers: If we're to highlight these as errors, let's ensure they're not highlighted when used as property names. The `throws` in `assert.throws`, for example, should not be considered an error.

syntax/javascript.vim
```diff
@@ -16,213 +16,193 @@ if !exists('g:javascript_conceal')
...
-
+" DISCUSS: Should we really be matching stuff like this?
 " DOM2 Objects
 syntax keyword jsGlobalObjects  DOMImplementation DocumentFragment Document Node NodeList NamedNodeMap CharacterData Attr Element Text Comment CDATASection DocumentType Notation Entity EntityReference ProcessingInstruction
 syntax keyword jsExceptions     DOMException
 
+" DISCUSS: Should we really be matching stuff like this?
 " DOM2 CONSTANT
 syntax keyword jsDomErrNo       INDEX_SIZE_ERR DOMSTRING_SIZE_ERR HIERARCHY_REQUEST_ERR WRONG_DOCUMENT_ERR INVALID_CHARACTER_ERR NO_DATA_ALLOWED_ERR NO_MODIFICATION_ALLOWED_ERR NOT_FOUND_ERR NOT_SUPPORTED_ERR INUSE_ATTRIBUTE_ERR INVALID_STATE_ERR SYNTAX_ERR INVALID_MODIFICATION_ERR NAMESPACE_ERR INVALID_ACCESS_ERR
 syntax keyword jsDomNodeConsts  ELEMENT_NODE ATTRIBUTE_NODE TEXT_NODE CDATA_SECTION_NODE ENTITY_REFERENCE_NODE ENTITY_NODE PROCESSING_INSTRUCTION_NODE COMMENT_NODE DOCUMENT_NODE DOCUMENT_TYPE_NODE DOCUMENT_FRAGMENT_NODE NOTATION_NODE
```
davidchambers: I don't take advantage of this matching, but it doesn't bother me.

syntax/javascript.vim
bounceme: these showing as warnings would help, but i think these should be detected by linters,unless we can make them errors only if used as identifiers

amadeus: I personally think we shouldn't even try to match or error correct these. I think leaving it up to a linter is a far better and more accurate solution. My vote would be to remove these completely since it's hard to figure what contexts they should and shouldn't be matched.

syntax/javascript.vim
amadeus: It only really bothers me from a 'keeping this syntax library clean' perspective, but I don't lose sleep over it.

syntax/javascript.vim
davidchambers: > I think leaving it up to a linter is a far better and more accurate solution.

Seems like a nice separation of concerns to me. :thumbsup:

daGrevis:
Bug in #393 seems to be fixed, good job!

Unfortunately I found a problem with HTML not being highlighted.

```javascript
class MyComponent extends React.Component {

    // HTML syntax works if I remove this.
    state = {
        isActive: false,
    }

    render() {
        return (
            <div>
                HTML doesn't have syntax highlighting. Syntax breaks on ' character.
            </div>
        )
    }

}
```

amadeus:
@daGrevis that's JSX syntax, which you can get by using this plugin [mxw/vim-jsx](https://github.com/mxw/vim-jsx)

fiedler:
I just tried the es6-refactor branch out on a few files in some of my projects and it seems to fix the issues with class instance fields and class static properties for me and I don't see any new problems.

It does - as you mentioned - require me to insert semicolons after object definitions. Would be great to find a solution for this but this is definitely a step up.

Thanks for working on this 👍

daGrevis:
> @daGrevis that's JSX syntax, which you can get proper syntax highlighting for using this plugin mxw/vim-jsx (it uses ours as a foundation, so it plays well).

Yes, I have that plugin already. Just saying that example above also breaks JSX highlighting.

> Also, what you are seeing is a very tricky issue to fix. If you place a semicolon after that object definition, it should correct the issue you are seeing.
>
> I am not sure there is valid/elegant way for me to fix that issue without requiring a semicolon

You are right, putting a semicolon at the end fixes the problem.

The thing is that me and a lot of other people are not using semicolons in JavaScript because they are optional. I'm not sure how I feel about cluttering the source code with them just to get correct syntax highlighting. Here's [a commit](https://github.com/daGrevis/svg-scene/commit/21f1d7bbb72ddb6b3f0dd62dfce2e9ed6d4ec1bb) where I added semicolons to a single file — quite a lot of noise. This also won't work for other people JavaScript code if they are not using semicolons.

What makes this issue so tricky to fix? Maybe we can put out heads together and try to fix it anyway? :)

Thanks for your work, I really appreciate it. :+1:

amadeus:
@daGrevis so it might be hard to give you a short answer as to why, but let me see what I can come up with here.

There's a high level discussion to be had for how accurate a syntax highlighter should be, but our philosophy has been that we want to catch obvious and simple syntax errors yet not get into more logical type problems that Flow, Tern or even ESLint would catch.

When creating a syntax regex in Vim, you have a few options - `match`, which takes a single regex and matches that to a `groupname`, `keyword`, which is basically a shorthand to match keywords without needing to write a regex, and finally the big one, `region`, which maps a specific region or scope based on a `start` and `end` regex.

Regions are special because they provide a new blank cope to work from that you can then opt other types of `region`s, `match`es or `keyword`s into. There is one caveat with `region`s however, as long as the starting regex matches, the region begins. It does not care if there is an ending region or not.

Something we do with this syntax file is create a ton of regexes that we then group into a special cluster called `jsExpression`. As the name implies, these are basically any sort of expression in JS. this include a variable, string, function, object, equations, etc. We pass this `cluster` into various regions as needed.

Example regions include an object (defined by a `{` and closing `}`), function definition, (which is defined by a `{` and `}` as well, however it's scoped to ONLY come AFTER the `(` and `)` of a function definition.

Using these scopes allows us, for example, to differentiate between an object key `someKey:` and a more generic ternary expression `test ? "true" : "false"`.

In the case of your example, we have a class definition block, also defined by `{` and `}`. Classes allow for special function type definitions like this (also, ES6 provides a similar shorthand syntax for objects):

```
class Example {
    // Inside here is a class definition region
    someFunc(){
        // Inside here is a function definition region
    }                   
}
```

This type of shorthand syntax is technically a syntax error if you put is in another context:

```
var func = someFunc(){};
```

In this case, someFunc() would be considered a function call of another function named `someFunc`, and the `{}` after it would technically evaluate to an object, this of course is a syntax error when it reaches the opening `{`.

Therefore, that specific regex to match a function definition, within a class block, is special, and isolated ONLY to the class blocks.

It gets complicated, because class definitions in ES7 syntax, support the ability to create expressions:

```js
class Example {
    someVal = 20
}
```

This is cool and all, and you're probably thinking, why don't we just throw the entire `jsExpression` cluster inside there. Well this becomes a problem in a case like this:

```js
var getValue = function(){
    return 2;
};
class Example {
    someVal = getValue()
    someFunc(){}
}
```

In this case, we need to match `getValue()` as a function call, but match `someFunc` as a function definition. If we just make `jsExpression` contained within the class block generically, then we will will run into a priority problem - whichever regex was defined later (function name or function call) will ALWAYS take precedence. So either you will ALWAYS get matched as a function call or a function definition.

The way to fix this is to create a new special region (caslled `jsClassValue`) that then follows an `=` sign WITHIN a class block that then contains the `jsExpression` cluster. This works great up until we have to figure out how to END that region. If you remember from earlier, the region will ALWAYS start, regardless of whether it finds an end or not. The `=` provides our start, however now we have to end that region. A semi colon makes it super easy, since we don't have to worry if newlines are involved, AND other contained regions inside of this new region will then extend the `jsClassValue` region conveniently. By this I mean if you did something like this:

```js
class Example {
    getValue = () => {
        return this._value;
    };
    someFunc(){}
}
```

The semicolon within the arrow method will not end the `jsClassValue` region. And it won't be until we end the function block region defined by the `{` and `}` that we match the closing `;` for `jsClassValue` region.

Javascript is introducing so many new fancy shorthand features, that it's becoming a huge nightmare to match using regexes.

Maybe others have some ideas?

amadeus:
Having said all that, I have just pushed a small change to add a newline `\n` as a region end, and I think this may actually work in most cases.

There is still the possibility of a sort of multiline statement breaking things, but I think at least for now, it could help.

Please pull the latest changes and test them when you have a chance.

daGrevis:
I can confirm that it works for every case in [svg-scene repo](https://github.com/daGrevis/svg-scene). I also tested it on my other private ES6 + JSX repo and there were no problems.

I will test it at work on Monday.

Thank you! :100:

amadeus:
Alright, I think this PR may be ready to go in. Obviously it's still for development, so there is more time to fix things. Any objections to getting this in now?

iansinnott:
Excellent work @amadeus! So far this is looking great

PR #470, opened 2016-06-12
JS Flow Annotation Support and A Proper Plugin System

amadeus:
I have begun work on JS Flow annotation support. This is a PR/Branch where it's currently on by default, however when this eventually gets merged, it should be hidden behind a global boolean, and perhaps even sourced from another file completely so as not to clutter up the main syntax file.

This is the test file I am using for it:

```js
function bar(x: string, y: number): string {
  return x.length * y;
}

function total(numbers: Array<number>) {
  var result = 0;
  for (var i = 0; i < numbers.length; i++) {
    result += numbers[i];
  }
  return result;
}

function foo(a: string, b: number): { key: number,  bool: boolean } {}

function foo(a: string, b: number): [ string, number ] {
}

function foo(a: string, b: number):object {}

function foo(a: string, b: number):object {
}

function foo(x: string): string { return x; }

function foo(a: string, b: number):void { }

var x: boolean = someBool;

class Bar {
  y: string;
  someMethod(a: number): string {
  }
}

let array: number[] = [1, 2, 3.14, 42];

let theAnswer: number = array[3]; // 42

let offTheEnd: number = array[100]; // No error

let array2: Array<string> = ["an alternate", "syntax", "for arrays"];

let tuple: [string, number, boolean] = ["foo", 0, true];

// Indexing into the array will return the type at a given index.
(tuple[0]: string);
(tuple[1]: string);

// Indexing into an statically unknown index will return a general type.
declare var unknownNumber: number;

// `void` is none of `string`, `number`, or `boolean`
(tuple[unknownNumber]: void);
(tuple[unknownNumber]: string|number|boolean); // OK

// Values written must be compatible with the type at that index.
tuple[1] = -1;
tuple[0] = false;

let object: {foo: string, bar: number} = {foo: "foo", bar: 0};
(object.foo: string);

function exampleMethod(someString: string, someRecord: Record): {resultString: string, resultString2: string} {}
```

I could probably use more test examples. Any of you who use JS Flow should please check out this branch and help test it!

Here's a screenshot of what JS Flow support currently looks like:

JS Flow Example


amadeus:
I've finally gotten around to adding a proper way for us to have plugins.

This particular PR adds plugins for Flow, NGDocs and JSDocs. I could probably still refine how this all works, for example there's a dependency issue where NGDocs is dependent on JSDocs, so I'll still need to come up with a system to standardize those dependencies.

bounceme:
how about setting the jsdoc variable in the ngdoc file?then users won't need to explicitly define it

amadeus:
Not sure I follow? NGDocs just add a few things to the existing JSDoc syntax. I don't think they should necessarily be on for JSDocs since they are very angular specific? Furthermore, if there are additional things people would like to add to NGDocs, then they could be added to that respective file.

bounceme:
What I was thinking is basically just adding a vimscript  `require('jsdoc')` at the top of the ngdoc file. i guess most people who want ngdoc will also already be using the normal jsdoc variable though

PR #472, opened 2016-06-13
Develop -> Master

amadeus:
Will need a version bump, me thinks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# vim-javascript v1.0.0
+# vim-javascript
 
 JavaScript bundle for vim, this bundle provides syntax highlighting and
 improved indentation.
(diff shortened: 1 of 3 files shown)


amadeus:
Also, I may need to rebase master back into develop, and fix some conflicts...

amadeus:
So I can do the rebase, but it will require a force push to `develop`.

How do you all feel about this?

bounceme:
I don't think thats a problem, i'm not too concerned with a perfect commit history

amadeus:
The only issue I see is people who might be tracking develop right now, and use a plugin manager. If I do a force push, then those people will have their develop branches freak out.

amadeus:
Another solution would be to still merge this in manually, but then create a new branch off of master to begin dev work on. But then that means people will stop getting updates on `develop`... Six of one, half a dozen of the other

bounceme:
a rebase sounds cleaner, most people will want to be on master now anyway

PR #475, opened 2016-06-14
Improvements to Destructuring - Needs More Testing

amadeus:
This PR should add far more complete support for destructuring, both in
var/let/const statements and also within function argument definitions.

There may be some cases I have not tested and are broken, therefore more
testing is for sure needed.

I've been using this file for reference:

```js
var [a, b] = [1, 2];
var [a, b, ...rest] = [1, 2, 3, 4, 5]
var {a, b} = {a:1, b:2}
var {a, b, ...rest} = {a:1, b:2, c:3, d:4};
var x = [1, 2, 3, 4, 5];
var x = [1, 2, 3, 4, 5];
var [y, z] = x;
var [one, two, three] = foo;
var [a, b] = [1, 2];
var [a=5, b=7] = [1];
var a = 1;
var b = 3;
const [a, b] = [b, a];
var a, b;
var [a, b] = f();
var [a, , b] = f();
var [, protocol, fullhost, fullpath] = parsedURL;
var o = {p: 42, q: true};
var {p, q} = o;
var a, b;
let {a, b} = {a:1, b:2};
var o = {p: 42, q: true};
var {p: foo, q: bar} = o;
var {a=10, b=5} = {a: 3};
function drawES6Chart({size = 'big', cords = { x: 0, y: 0 }, radius = 25} = {}){}
const { Loader, main } = require('toolkit/loader');
var { title: englishTitle, translations: [{ title: localeTitle }] } = metadata;
for (var {name: n, family: { father: f } } of people) {}
function userId({id}) {}
function whois({displayName: displayName, fullName: {firstName: name}}){}
let { [key]: foo } = { z: "bar" };
```

And for reference, what it looks like using my custom colorscheme:

destructuring example

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -369,7 +369,8 @@ function GetJavascriptIndent()
 
   " If line starts with an operator...
   if (line =~ s:operator_first)
-    if (s:Match(lnum, s:operator_first) || s:Match(lnum, s:line_pre . '[])}]'))
+    if (s:Match(lnum, s:operator_first) || (s:Match(lnum, s:line_pre . '[])}]') &&
+          \ !(s:Match(v:lnum,s:line_pre . '\.') && s:Match(lnum, ')' . s:line_term))))
       " and so does previous line, don't indent
       return indent(lnum)
     end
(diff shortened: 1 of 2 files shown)


amadeus:
This should be opened against `develop`

PR #479, opened 2016-06-14
Adds support for string func names in objs classes

### LINKED ISSUE #477: Quoted string as object key not supported
Since pulling the updates from today, my highlighting has been way off.

screen shot 2016-06-14 at 10 37 39 am

None of the `const`s are highlighted anymore, as are... Well, most other things.  It was all fine before `develop` was merged in today.

***

**Edit:** It's been determined that the issue is specifically with the following kind of syntax

```js
const obj = {
  'my-method'() {
    return ...;
  }
};
```

amadeus:
This fixes #477

Basically this adds support for function definitions like this:

```js
var obj = {
    "some-func"(){},
    'another-func'(){}
};

class Example {
    "some-func"(){},
    'another-func'(){}
};
```


amadeus:
This should definitely get merged into `develop` and into `master` asap. It's a simple and very important fix.

PR #489, opened 2016-06-17
incorrect change in cursor pos

bounceme:
some bizarre problem which made me invent a buggy fix. also reindented the file


bounceme:
i'll add some comments to the functions that change the cursor's pos. this hopefully gets in master soon

bounceme:
also fixes #492

PR #499, opened 2016-06-19
accurate switch indenting

bounceme:
since cindent considers single quotes as delimiting a single char string, cindent will break in situations like : https://github.com/sanctuary-js/sanctuary-def/blob/master/index.js#L222

http://stackoverflow.com/questions/3683602/single-quotes-vs-double-quotes-in-c


bounceme:
do let me know if there is a better way to make changes to `cpo`, but everything is working fine

Issue #497, opened 2016-06-19
jsFuncArgRest break if there's no space before closing parentheses


Arch-Mage:
With this

```javascript
function something(arg1, ...rest) {
   return;
}
```
that closing parenthesis will be highlighted by `jsFuncArgRest`.

but with this:

```javascript
function something(arg1, ...rest ) { // notice space after 'rest'
   return;
}
```

that parenthesis is become `jsFuncParens`, which I think should be.

If I'm not mistaken, maybe the causing problem is the regular expression in [this line](https://github.com/pangloss/vim-javascript/blob/master/syntax/javascript.vim#L151):

```vim
/\%(\.\.\.[a-zA-Z_$][0-9a-zA-Z_$]*\))/
```

I've tried to remove one parenthesis in that double parenthesis, so it become like this:

```vim
/\%(\.\.\.[a-zA-Z_$][0-9a-zA-Z_$]*\)/
```

And it works.

amadeus:
I've actually just reworked all of this in the `develop` branch, can you check it out? It should be fixed.

Arch-Mage:
Ah, I didn't know that. Yes, the `develop` branch fix this issue. Oh, and I found another issue. The current master branch (and current develop branch as I see there) breaks [Export extensions transform](http://babeljs.io/docs/plugins/transform-export-extensions/).

amadeus:
What exactly is breaking? This should be working fine?

amadeus:
Nevermind, I see the issue. I'll look to add this soon.

amadeus:
@Arch-Mage since this original issue is in regards to an issue that has been fixed, and I have created a PR (#500) for the other feature request, I am going to close this.

PR #500, opened 2016-06-20
First pass at improved export support

amadeus:
A first pass at improving `export` support, via this [comment](https://github.com/pangloss/vim-javascript/issues/497#issuecomment-227011774)


amadeus:
Gonna get this is now, so we can get a `develop` => `master` PR ready and tested.

PR #504, opened 2016-06-21
jsComment updates

amadeus:
Removed some seemingly needless regexes, updated to match consistency.

@bounceme Can you see if this branch fixes anything with the comment stuff you mentioned?


bounceme:
i think this fixes those issues

amadeus:
Oh interesting, I wonder if it was due to this weird regex:

```
syntax region  jsComment        start=+^\s*\/\/+ skip=+\n\s*\/\/+ end=+$+ contains=jsCommentTodo,@Spell fold extend keepend
```

I wasn't sure what in the hell it was supposed to do... Oddly specific with a strange skip mechanism. Seemed like a recipe for bugs.

PR #515, opened 2016-06-28
indentation refactoring

bounceme:


bounceme:
i would like some testers for this, everything is working(except for a few string problems,and generally worse performance, which i hope to fix)

bounceme:
this is largely finished, if your reading this can you please test? performance suffers within large blocks of code but the simplification makes this a fair trade in my opinion

markrian:
I'll test this today, and report here with any regressions or bugs.

markrian:
Testing has so far revealed only one thing I'd definitely call a bug, and a couple of other things which are perhaps a matter of personal preference. The former I can only reproduce on a fairly large file, and I'm having trouble producing a minimal test case. In fact, it almost seems like the size is to blame, since changing the file at all seems to resolve the bug. That file is proprietary, so I can't share it, sadly. Once I've managed to make a minimal test case, I'll put it here.

I'll describe the other personal preference "bugs" later, too, when I have some time.

markrian:
Bizarrely, it seems that this bug I mentioned only happens when the file I'm working with is 1035 lines or more long. If I remove *any* line in the file, even blank lines, before or *after* the incorrectly indented code, it then indents the whole file correctly.

bounceme:
these are caused by the timeout for the search pair function, default is 300 ms. maybe it should be configurable if this gets merged

bounceme:
i have a few ideas for improving the speed issue, 300 ms is too much(though it would only take that long in a very big block scope)

bounceme:
@markrian there have been improvements to this, is this working any better?

markrian:
I'll take a look tomorrow. Sounds good!

markrian:
My testing so far has revealed two bugs.

At no indentation, any lines after the first statement are incorrectly indented:
```javascript
a();
    b();
    function x() {}
```
And multiline strings can mess up subsequent indentation:
```javascript
function a() {
    a(`
    `);
        b();
        c();
    }
```
removing the newline in the string fixes it:
```javascript
function a() {
    a(`    `);
    b();
    c();
}
```
Other than these two issues, it seems solid! Although it is still a little slow, perhaps.

bounceme:
I did just fix a template string issue though not the specific one you are showing. is it reproducible on develop?i can't reproduce your first issue, though i have seen that behavior on earlier commits, are you fully updated?

bounceme:
also if the performance can't be improved further I don't know if this should even be merged,though i wouldn't want to deal with the complexity of the current script anymore

bounceme:
that other problem should be fixed now

markrian:
Hmm, the initial indent problem seems fixed initially, but subsequent indentation calls break it. So this text, after one `gg=G` is indented correctly:
```javascript
a();
b();
function x() {}

function a() {
}
```
but do `gg=G` again and you get this:
```javascript
a();
    b();
    function x() {}

    function a() {
    }
```
and again and you get this:
```javascript
a();
        b();
        function x() {}

        function a() {
        }
```
I can reproduce the template string issue on develop as well. In both develop and indent-refactor, however, the problem only appears when you tell vim to indent the lines explicitly, using the `=` command. While typing, the correct indentation is used.

This branch is still about twice as slow as develop. On a particularly nasty 1035 line file I have, this branch takes more than 8 and a half seconds to indent it:
```
$ time vim -c "normal gg=G" -c 'q!' nasty.js 

real	0m8.541s
```
while develop does it in just over 4 seconds
```
$ time vim -c "normal gg=G" -c 'q!' nasty.js 

real	0m4.311s
```
That said, I'd happily accept that performance hit for more robust indenting!

bounceme:
Hey, hopefully the last commit fixes that bug, also compare the performance and see if this is an improvement,thanks!

bounceme:
i've done something wrong, don't bother testing yet

markrian:
```
$ time vim -c "normal gg=G" -c 'q!' nasty.js

real	0m3.317s
```
Nice!

You'll also be pleased to hear that I can't find any bugs that aren't in develop - great work!

amadeus:
Sorry I haven't had time to test this, I've been pretty busy with work.

@markrian what's the time on the default vim as a reference for this?

bounceme:
Know that this will still be slower than the develop branch in certain files. Mainly files with very large "shallow" scopes which are also parentheses dense

markrian:
@amadeus, develop indents this file in ~4.3s. This branch now does it in ~3.3s. Vim's standard JS indenter, however, manages it in ~0.3s.

amadeus:
So this sounds like it might be an improvement over `develop` then. I'll try and test it out soon.

bounceme:
I think this is about done, if you see any regressions let me know. also i am glad to hear any suggestions on what i've tried here

bounceme:
I'll merge later today or tomorrow if its alright with @amadeus & @markrian

markrian:
I'm afraid I've found another indenting bug. Trying to produce minimal test case.

markrian:
It seems to be another idempotency issue. The follow correctly indented code stays intact after one `gg=G`:
```javascript
'use strict';

function foo() {
}
```
but a second call results in this:
```javascript
'use strict';

    function foo() {
    }
```
and subsequent calls indent the function further and further.

markrian:
Just found another one. Do/while loops mess up indentation of the line immediately following them:
```javascript
function foo() {
    do {
        anything();
    } while (false)

         anything();
    anything();
}
```

bounceme:
i think the first issue is fixed but the do while syntax makes it very unlikely that can be solved unless there is a do block syntax element added, which may or may not be possible

bounceme:
actually,i'll try and port this
https://github.com/mooz/js2-mode/blob/master/js2-old-indent.el#L253

bounceme:
been on vacation this past week, hopefully someone was testing this as I'd like to finish and merge it tomorrow

amadeus:
Overall, things have been pretty good, however I have found a couple weird instances, mostly working within JSX files, not sure if they are related to JSX, or if it's something deeper with JS.

I'll see if I can isolate the problems over the next couple days.

amadeus:
However, at this point, it may be worth merging this in and then working to fix the rest of the issues over time? Also getting it into `develop` will inevitably get more eyes on it and therefore more issues may be found to get fixed.

bounceme:
Sounds fine, i'll merge after a few finishing touches

Issue #517, opened 2016-06-28
Flow type value missing semi-colon terminator


gf3:
Example snippet:

``` js
type State = {
  errorMessages: Array<string>;
  forceValidation: bool;
  invalidFields: Array<string>;
  submitting: bool;
};


//-----------------------------------------------------------------------------
// Blah blah
//-----------------------------------------------------------------------------

function lol() {
   // ...
}
```

can fix by changing:

``` vim
syntax region  jsFlowTypeValue      contained matchgroup=jsFlowNoise start=/:/       end=/[,}]/ contains=@jsFlowCluster
```

to

``` vim
syntax region  jsFlowTypeValue      contained matchgroup=jsFlowNoise start=/:/       end=/[,};]/ contains=@jsFlowCluster
```

amadeus:
#518 should fix this problem. But that branch also includes some bigger fixes as well, so it would be helpful for you to test that branch and see if I didn't cause any other inadvertent issues.

gf3:
tested and fixed in #518

amadeus:
Fixed in `develop`

PR #518, opened 2016-06-28
More flow fixes

### LINKED ISSUE #516: 'Maybe' flow types break syntax highlighting
On v1.1.3
Optional flow types (e.g. var foo: ?string = 'bar') will break the highlighting on subsequent lines.

Same result with both `let g:javascript_plugin_flow = 1` and `let g:javascript_plugin_flow = 0`

screen shot 2016-06-28 at 14 15 27

amadeus:
A series of fixes for JS Flow.

Fixes #516 and #517 as well as a couple other small things that I could in my own tests.


amadeus:
Some pretty important flow fixes, since this is a plugin, don't really think any serious approval is in order

PR #531, opened 2016-07-11
Remove jsOperator from jsClassBlock

jridgewell:
It screws up highlighting of the operators when they are used as method names:

`delete` keyword as class method.

This doesn't happen if you're inside a normal object:

`delete` keyword as obj method


amadeus:
Unfortunately we can't do that or we break all of our ES7 class expression support.

I'll look at isolating delete out of `jsOperators` so it doesn't break classes shortly.

jridgewell:
> Unfortunately we can't do that or we break all of our ES7 class expression support.

I'll defer to you, but nothing seems broken yet.

> I'll look at isolating delete out of jsOperators so it doesn't break classes shortly.

It's not just delete, all of the `jsOperators` are valid method names. `delete` just happens to be the most annoying since it's being used in a lot of the newer native classes.

amadeus:
this is the kind of stuff that would break https://github.com/jeffmo/es-class-fields-and-static-properties/blob/master/README.md

amadeus:
Actually, you know what, I just realized this is totally ok because we actually use a region to define those expressions and therefore it should actually be fine.

jridgewell:
Whoot! Now if only I could make `this.delete()` not highlight weirdly.

amadeus:
Yeah, that one is a bit tougher to to fix. Right now we rely on `syntax keyword` for those, but we'd have to change it to a regex that looks ahead or behind, unfortunately :(. Theoretically this could hurt performance in large files as well.

Issue #487, opened 2016-06-16
Misc Issues With Strangely Placed Comments


amadeus:
I have found a few cases where placing comments in specific contexts can cause major problems when relying on `nextgroup`.

Some examples:

```js
if (someBool) // This is a comment
{
   // This region is now matched as a `jsObject`, which is incorrect.
}

switch(someStr) /* Another comment */ {
    // This region is now matched as a `jsObject`, which is incorrect.
}
```

I wish there was a way to classify a certain group of regexes as skippable, and then opt in to skipping over them for determining `nextgroup`, however unfortunately this isn't possible.

The only real fix is to create case specific comments that can act as go-betweens for these regions. Since they would be contained I don't think it would create much overhead, but it would fix.

I am creating this issue as a reminder to fix these cases.

smiechowy:
Yup, this is especially visible when using flow comment annotations
```
function testFunc() /*: string*/{
    var testObj = {
        prop: "lorem ipsum"
    };

    return someOtherFunc(testVar.prop);
}
```
highlight after /*: string*/ is totally broken

amadeus:
@smiechowy we've recently added some experimental flow annotation support. Check the readme for details. It won't fix your comment annotation but it should support uncommented annotations

amadeus:
Since you've brought up this issue, I will look into adding fixes for these comments in the next day or so.

smiechowy:
Unfortunately I'm not using uncommented annotations in my current project, but one day it might come in handy. 

> Since you've brought up this issue, I will look into adding fixes for these comments in the next day or so.

Awesome, I'm looking forward to some results. Thanks for your hard work :)

amadeus:
Took me a couple days to find time to tackle this, however I have a PR ( #532 ) open against develop that should fix many common instances of these.

amadeus:
This issue should be mostly fixed in `develop`

PR #532, opened 2016-07-11
Adding support for oddly placed comments

amadeus:
I have added support for a few basic comment positions. So far this is not really meant to be an exhaustive list of all possibilities, but really the most common patterns, and we can continue to add other comment patterns as people raise issues for them.

Here is a small snippet showing the ones that have been fixed.

```js
class Durp /**/ {
}

class Durp // durp
{
}

class Durp extends Tube // durp
{
}

function testFunc() /**/ {
}

function testFunc() // durp
{
}

if (true) // asdlkfjas d
{
  var durp = 2;
} else /* durp */
{
  var durp = 4;
}

while (true)  // durp
{
  var durp = 2
}

for () /* durp */
{
  var durp = 2;
}
```

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -95,7 +95,8 @@ exe 'syntax keyword jsSuper     super  contained '.(exists('g:javascript_conceal
 
 " Statement Keywords
 syntax keyword jsStatement    contained break continue with yield debugger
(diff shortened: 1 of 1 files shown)


amadeus:
This should fix #487

PR #533, opened 2016-07-13
typo

romgrk:
Additionnal thoughts: 
- thought about having NodeJS keywords?
- Add `@Comment` group to `syntax region  jsComment        start=/\/\// end=/$/ contains=jsCommentTodo,@Spell extend keepend`? That way people can add their own comment-highlighting whitout hacking through the whole syntax.


amadeus:
Thanks for submitting the PR, good eye.

Since NodeJS is so common, I am totally ok with adding certain important NodeJS keywords. Other NodeJS related stuff could always be added in a plugin as well.

The `@Comment` group is not a bad idea, I can certainly look into something like this soon.

PR #535, opened 2016-07-15
Tweak highlighting of numeric literals

dmitrivereshchagin:
- Add highlighting for binary and octal number syntax
- Highlight unary minus as operator regardless of `'iskeyword'` option
- Remove `L` from integer literals


bounceme:
Thank you for your contributions! would be best to have these merge into dev though

dmitrivereshchagin:
Sorry. Is it possible to update pull request to merge into another branch?

bounceme:
i don't think so, you can just create a new p.r.

PR #540, opened 2016-07-15
Tweak highlighting of numeric literals

dmitrivereshchagin:
- Add highlighting for binary and octal number syntax
- Highlight unary minus as operator regardless of `'iskeyword'` option
- Remove `L` from integer literals


amadeus:
So unfortunately these PRs you've opened need to based off of, and against the `develop` branch. I am curious to test them out though, so sorry for the confusion here.

amadeus:
Oh shoot, I think you fixed this, I am on mobile which doesn't show the base branch, and I just remember seeing these from earlier

amadeus:
I'll test this PR out tomorrow at work. I think it should be ok, but I just want to see it in action first.

PR #541, opened 2016-07-15
Make use of syntax specific iskeyword setting

dmitrivereshchagin:
It's better not to touch `'iskeyword'` option from a syntax file if possible.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -16,8 +16,12 @@ if !exists('g:javascript_conceal')
   let g:javascript_conceal = 0
 endif
 
-" Dollar sign is permittd anywhere in an identifier
-setlocal iskeyword+=$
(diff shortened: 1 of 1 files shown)


bounceme:
I think we should set iskeyword regardless, in the ftplugin file, because it changes vim movements to not stop on '$' in identifiers

dmitrivereshchagin:
There's nothing wrong with setting `'iskeyword'` from filetype plugin too. Technically it would do the same thing but serve different purpose. I think that syntax files shouldn't affect movements and searching. There's nothing we can do with older Vim versions but at the time `syntax iskeyword` is recommended by Vim builtin documentation.

amadeus:
I like and approve this change.

bounceme:
i'm fine with this, thanks @dmitrivereshchagin  !

PR #543, opened 2016-07-15
Update filetype plugin

dmitrivereshchagin:
- Add header
- Replace `'re'` with `'regexpengine'`
- Simplify patch check

### DIFF
--- a/ftplugin/javascript.vim
+++ b/ftplugin/javascript.vim
@@ -1,4 +1,10 @@
+" Vim filetype plugin file
+" Language:     JavaScript
+" Maintainer:   vim-javascript community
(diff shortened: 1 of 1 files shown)


bounceme:
should we add a `undo_ftplugin` variable?

bounceme:
maybe also make the iskeyword setting be set in the ftplugin if the vim version is too old

dmitrivereshchagin:
Yes we should. But to make it work one of the following should be done

- provide full filetype plugin which set `b:did_ftplugin` variable
- move current filetype plugin to `after/ftplugin` directory

In other cases filetype plugin bundled with Vim will override `b:undo_ftplugin`.

Issue #524, opened 2016-06-29
Incorrect indentation after multiline variable declaration with array or object


markrian:
Tested on latest develop ad5d09edfff3b15ac1c785796abb1705a0d64af6. This is the resulting indentation:
```javascript
const a = 1,
    b = {
        foo: 'bar'
    };

    anythingHere();
```
whereas I would expect this:
```javascript
var a = 1,
    b = {
        foo: 'bar'
    };

anythingHere();
```
Bizarrely, a trailing comma in the object or array literal fixes the subsequent indentation:
```javascript
const a = 1,
    b = {
        foo: 'bar',
    };

anythingHere();
```

bounceme:
my new p.r. #515  , if finished, will work in just about any situation regardless of keyword, comma etc. it also removes about 80% of the code

markrian:
Testing has so far revealed only one thing I'd definitely call a bug, and a couple of other things which are perhaps a matter of personal preference. The former I can only reproduce on a fairly large file, and I'm having trouble producing a minimal test case. In fact, it almost seems like the size is to blame, since changing the file at all seems to resolve the bug. That file is proprietary, so I can't share it, sadly. Once I've managed to make a minimal test case, I'll put it here.

I'll describe the other personal preference "bugs" later, too, when I have some time.

amadeus:
This is now fixed in `develop`

amadeus:
Should now be fixed in `develop`

PR #544, opened 2016-07-15
Develop -> Master

amadeus:
Still some more work to be done, and I have some JSX improvements I'd like to get in here soon, but getting the PR started to kick us into gear to get these changes into `master`

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,24 +35,37 @@ And install it:
 
 ## Configuration Variables
 
-The following variables control certain syntax highlighting features. You can
-add them to your `.vimrc` to enable/disable their features.
+The following variables control certain syntax highlighting plugins. You can
+add them to your `.vimrc` to enable their features.
+
+-----------------
 
 ```
-let g:javascript_enable_domhtmlcss = 1
+let g:javascript_plugin_jsdoc = 1
 ```
 
-Enables HTML/CSS syntax highlighting in your JavaScript file.
+Enables syntax highlighting for [JSDocs](http://usejsdoc.org/).
 
 Default Value: 0
 
 -----------------
 
 ```
-let g:javascript_ignore_javaScriptdoc = 1
+let g:javascript_plugin_ngdoc = 1
 ```
 
-Disables JSDoc syntax highlighting
+Enables some additional syntax highlighting for NGDocs. Requires JSDoc plugin
+to be enabled as well.
+
+Default Value: 0
+
+-----------------
+
+```
+let g:javascript_plugin_flow = 1
+```
+
+Enables syntax highlighting for [Flow](https://flowtype.org/).
 
 Default Value: 0
 
--- a/after/ftplugin/javascript.vim
+++ b/after/ftplugin/javascript.vim
@@ -3,8 +3,6 @@
 " Maintainer:   vim-javascript community
 " URL:          https://github.com/pangloss/vim-javascript
 
-setlocal suffixesadd+=.js
+setlocal iskeyword+=$ suffixesadd+=.js
 
-if v:version == 703 && exists('&regexpengine') || v:version == 704 && !has('patch2')
-  set regexpengine=1
-endif
+let b:undo_ftplugin .= ' | setlocal iskeyword< suffixesadd<'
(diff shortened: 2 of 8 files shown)


bounceme:
i have to work on the switch block indentation before a merge, hope to finish soon

amadeus:
No worries, I want to re-look into some JS Flow stuff this weekend before we merge into `master`

bounceme:
So i'm having some pretty serious bugs with the indentation, do you mind if i go back to the master version of the script? also any suggestions on the best way to revert, should I just make a new commit or try and revert?

bounceme:
edited ^

amadeus:
@bounceme what was the problem with the refactored indent script? It appeared to be working pretty well?

bounceme:
i thought so also, then i tested with this file  http://pastebin.com/wrDcRTcd
also there are some strange inconsistencies with the performance technique i used to determine the type of pair to look for with the lookforparens function

bounceme:
alright, looks like i fixed the refactored version potentially,a revert would be easy,and would restore the history(though i want to take better care to make sure we should before i do)

bounceme:
i guess it is a matter of preference between current dev and https://github.com/pangloss/vim-javascript/pull/552 with some improvements. anyone have strong feelings on which we should use?

amadeus:
@bounceme not sure I follow, using current `develop` #524 seems fixed?

bounceme:
sorry, basically the version of the script here: https://github.com/pangloss/vim-javascript/blob/a7d5f9f83dcd2ab7da8e10cbba52713f8ff8f116/indent/javascript.vim (524 is fixed though there are a few bugs not related to that,which are fixable) and the newer,smaller,slower(in general) version that is in develop

amadeus:
My thoughts thus far:

1. You say the new indent script is slower, yet based on this [comment](https://github.com/pangloss/vim-javascript/pull/515#issuecomment-230848418), it's comparable or even faster. Furthermore I have NOT noticed any performance hits in my day to day usage of it.
2. My assumption is this new version of the script is a lot easier to work with and smaller?

To me, both those points would qualify as a major win and definitely worth merging in.

You mentioned there may be a couple small issues, would you mind detailing them here or in a separate issue so we can be aware of them/have a log of them? Would those issues be fairly solvable in the future?

bounceme:
edited

bounceme:
we should probably just use the new version of the indentation

bounceme:
also we can't easily support things like this since everything is based on offsets inferred by braces
```
  if()
    if()
      if()
        console.log("value");
```

bounceme:
unrelated but, can we do something about the gq code in this pr?

amadeus:
@bounceme regarding your first comment on the if statements, not sure what exactly you changed in that patch, but it does work rather nicely. However I haven't had time to test if there are other issues it may introduce?

In regards to `gq`, what kinds of changes would you like to address?

bounceme:
so i think that also works in master, haven't checked though. About the gq stuff, maybe we can get some input from users about whether they'd like to have it included here or not.i think it should be removed

amadeus:
Right now, it appears that `gq` just puts everything on the same line, essentially the sames as selecting a bunch of content and pressing `shift + j`.

If this is the case, `gq` should probably be removed

bounceme:
indent script is ready! how is the jsx/flow work going @amadeus ?

amadeus:
Alright, I think this is ready to go. @bounceme any last things you might want to do?

bounceme:
Not really, this is ready

amadeus:
@bounceme what about the PR you just opened? Should we wait for that?

bounceme:
No ,its just a weird edge case
these issues are present currently:
```
if
()
fun()
```
and continuations within 'c-style' 
```
function ()
{
  var c = 1,
  o= r
}

bounceme:
Maybe I could just merge it now though? only the first example is fixed, but having it tested by users be good

amadeus:
@bounceme yeah that's good with me, go for it!

bounceme:
you do the honors @amadeus !

bounceme:
actually I have a possible fix for the 'c-style' issue

PR #548, opened 2016-07-15
for 194

bounceme:


amadeus:
@bounceme this looks good to me! Just depends on how confident you are about it

Issue #423, opened 2016-05-08
Indentation of switch case inside a callback


eddiezane:
On `develop`.

```javascript
app.post('/events', function (req, res) {
  switch (req.body.foo) {
  case 'participant-join': // not indented far enough
    break // not indented far enough
    case 'participant-leave':
      delete db[req.body.CallSid]
      break
  }
})
```

<img width="519" alt="screen shot 2016-05-08 at 12 36 20 pm" src="https://cloud.githubusercontent.com/assets/959573/15099027/7ff961be-1519-11e6-8985-13b5f1802591.png">

bounceme:
I see some problems with indenting with your example, but not actually the same as you, check to see what your cinoptions are, https://github.com/pangloss/vim-javascript/pull/400#issuecomment-214596214

bounceme:
oh never mind, i also see this

bounceme:
I suspect this is fixed!

eddiezane:
Thanks @bounceme! You're crazy fast on these 😄

PR #551, opened 2016-07-17
Remove unused options from indent plugin

dmitrivereshchagin:
When `'indentexpr'` is set, setting `'smatrindent'` or `'cinoptions'` has no effect.

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -11,15 +11,13 @@ if exists("b:did_indent")
 endif
 let b:did_indent = 1
 
-setlocal nosmartindent
-
 " Now, set up our indentation expression and keys that trigger it.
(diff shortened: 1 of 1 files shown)


bounceme:
make sure #423 still works

bounceme:
we use c-indent in the indentexpr

dmitrivereshchagin:
Oops. Haven't noticed that. Leave `'cinoptions'` alone.

bounceme:
is `set nolisp ` necessary? https://github.com/neovim/neovim/search?utf8=✓&q=nolisp

dmitrivereshchagin:
Setting `'lisp'` can break indenting. But I think it's very unlikely that someone sets it.

Issue #93, opened 2013-07-29
Regular expressions /\[/ and /\]/ break syntax highlighting


moll:
For example:

``` javascript
.replace(/\[/g, ".").replace(/\]/g, "")
```

Results in the following redness:
![Broken syntax highlighting](https://s3.amazonaws.com/f.cl.ly/items/0K2Z2k2b263j0n0V2B0Y/Screen%20Shot%202013-07-29%20at%202.15.33%20PM.png)

Issue #88, opened 2013-07-23
Regular expression and quote (/"/) breaks syntax highlighting


Rob--W:
Steps to reproduce:

```
var x = /"/;
var y = '';
```

Output:

![bad](https://f.cloud.github.com/assets/1365071/840898/321fe88a-f381-11e2-998c-3e765a9d2268.png)

Expected output (colors are different, but at least the highlighting is consistent):
![good](https://f.cloud.github.com/assets/1365071/840903/51288476-f381-11e2-9dc6-8a2c1d0fbe99.png)

PR #553, opened 2016-07-17
Update ftplugin

dmitrivereshchagin:
Removed `'regexpengine'` workaround. Setting `'regexpengine'` from filetype plugin without reenabling syntax doesn't fix problems described in #88 and #93.

I have tested some old versions of Vim. Seems like those problems are reproducible only in versions 7.4 and 7.4.1. Those few unfortunates who encounter it today will be able to find fix in project issues.

Filetype plugin moved to after directory so now it's possible to use `b:undo_ftplugin` variable.

### DIFF
--- a/after/ftplugin/javascript.vim
+++ b/after/ftplugin/javascript.vim
@@ -3,4 +3,6 @@
 " Maintainer:   vim-javascript community
 " URL:          https://github.com/pangloss/vim-javascript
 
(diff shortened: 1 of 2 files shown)


bounceme:
looks fine, thanks

bounceme:
could you add `iskeyword` please?

dmitrivereshchagin:
Updated.

PR #560, opened 2016-07-20
Update indent plugin

dmitrivereshchagin:
- Restore `'cpoptions'` at the end of the file so it affects all functions
- Rename `Feixedgq()` to `GetJavascriptFormat()`

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -206,9 +206,6 @@ function GetJavascriptIndent()
 
 endfunction
 
-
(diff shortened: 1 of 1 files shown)


bounceme:
https://github.com/pangloss/vim-javascript/pull/544#issuecomment-233865792
Looks like we'll jus

dmitrivereshchagin:
I agree with removing `'formatexpr'` setting. Closed.

Issue #559, opened 2016-07-20
option for extra indent in chained method calls


gera2ld:
For example:
``` js
// Expected, situation in Atom
Promise
.resolve()
.then(() => ())
```

In Atom, it will be indented just like above. But with vim-javascript, there will be an extra level of indent before each method:
``` js
// Actual situation
Promise
  .resolve()
  .then(() => ())
```

Sometimes there will be lots of chained calls. I don't think it necessary to indent since we don't actually care about the start of the chain and we can add blank lines to separate different blocks.

Perhaps some peopl prefer the extra-indented way, so can we have an option for this?

bounceme:
Alright, here's an exercise for the people requesting this : make a p.r. and I'll merge if it works correctly and leaves the current behavior as default

amadeus:
So, I partially agree with the request in this issue, however I think there is a more balanced middle ground that doesn't require configuration options.

As of now, the default functionality does exactly the first scenario:

```js
Promise
.resolve()
.then(() => ())
```

While I personally would use the desired results in the follow-up example:

```js
Promise
  .resolve()
  .then(() => ())
```

I think it would get annoying for people who like to write JS without semicolons (I am not one of those people btw :P). However, I think a preferred middle ground is the following (although this may not be possible, I am not sure?):

```js
Promise
|
^ - Cursor by default goes here
```

```js
Promise
    |
    ^ - I press tab to indent, and type the next method
```

```js
Promise
    .resolve()
|
^ - When I press return, I am brought back to the original indentation
```

```js
Promise
    .resolve()
    |
    ^ - is it possible to be returned to here?
```

I guess I have no idea how to know when the user is done with a statement or when they continue a statement... so not sure how solvable this is?

bounceme:
@amadeus we can experiment with `indentkeys`
try this : `set indentkeys+=0.`

bounceme:
is that configurable in atom?

gera2ld:
@bounceme it's not configurable by default but can be supported by plugins in atom.
Some people are requesting for this feature too (beautify-web/js-beautify#482).

bounceme:
#566

PR #563, opened 2016-07-20
feat: add config for indenting chained method calls

gera2ld:
- add g:javascript_plugin_chain_indent
- See pangloss/vim-javascript#559

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,16 @@ Default Value: 0
 -----------------
 
 ```
+let g:javascript_plugin_chain_indent = 0
+```
+
+Disables extra indent for chained methods.
+
+Default Value: 1
+
+-----------------
+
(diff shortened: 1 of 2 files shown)


bounceme:
thank you, there isn't really any point in merging this now since there may be large changes to the indent script soon. I think we should just make the `s:operator_first` and `s:continuation_regex` variables global

bounceme:
will that be alright ? making the regexes global so it is completely configurable and we don't need global options for every operator (which I am against)

gera2ld:
Sounds good. Do I need to update the PR or just close it?

bounceme:
you can close it, I'll do this before we merge dev -> master

bounceme:
#566

PR #595, opened 2016-08-10
remove syntax check

bounceme:
this seems to work so far, only thing that i have to address is multiline strings and  template literals
oh and regex
credit to https://github.com/jiangmiao/simple-javascript-indenter


bounceme:
i may reopen later, https://github.com/sweet-js/sweet.js/wiki/design could probably work with the `s:Onescope` function

bounceme:
Actually, i think template strings invalidate this approach to lexing, since the start and ending delimiter is completely ambiguous without parsing from the top of the file down

PR #596, opened 2016-08-12
block statement

bounceme:
shouldn't be too difficult

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -79,6 +79,12 @@ function s:Onescope(lnum,text,add)
         \ (a:add || (expand("<cword>") ==# 'while' ? !s:lookForParens('\<do\>\C', '\<while\>\C','bW',100) : 1))
 endfunction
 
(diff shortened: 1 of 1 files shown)


bounceme:
So this works fine, it's just really slow right now and i have no idea why. also this is what this pr changes:
```
{
  var i = 0,
    l = 5
}
// vs
{
  var i = 0,
  l = 5
}
```

bounceme:
so there is a slight decrease in performance with this pr however this uses the same algorithm from https://github.com/sweet-js/sweet.js/wiki/design#give-lookbehind-to-the-reader which the acorn parser also uses. it actually is basically the same as what i had originally but with support for labeled block statements and plain block statements.

My results with https://github.com/sanctuary-js/sanctuary-def/blob/master/index.js

`block : 10.327 seconds`
`dev : 8.689 seconds`

bounceme:
will probably merge this tomorrow evening after some more testing

PR #601, opened 2016-08-16
cleaning

bounceme:


bounceme:
Hey @amadeus, what do you think about a dev->master push? I'd like to get the new indent changes out

amadeus:
definitely get the PR for it open, and we can begin making sure tested changes are in it.

Issue #598, opened 2016-08-13
async keyword not properly highlighted in various places


Arch-Mage:
I found `async` keyword is not properly highlighted places. Here some which I found.

```javascript
export default async function fn() {} // Got 'jsExportContainer'
export async function fn() {} // Got 'jsExportContainer'

class Class {
  async method() {}  // Got 'jsClassMethodType'
  static staticMethod = async function() {} // Got 'jsClassValue'
}

const object = {
  async method1() { }, // Got 'jsObjectMethodType'
  method2: async function() { }, // Got 'jsObjectValue'
  method3: async () => {} // Got 'jsFuncCall'
}
```
I think `async` keyword there should be highlighted as `jsAsyncKeyword`.

mixedCase:
@amadeus Voting for reopening. Bug still present for the most part. Copypasting OP's snippet gives me this: https://i.imgur.com/fifGcOp.png

leomao:
Is this issue going to be fixed?
Adding `jsAsyncKeyword` into `jsExpression` cluster can solve this.
But I am not sure if it is the correct way to solve this problem...

PR #602, opened 2016-08-16
fix async keyword and comment inside destructuring

Arch-Mage:
I think `async` keyword should be an optional part of
a function. So, I remove `async` from class and object
method type and move it to be an optional part in
function regex. This is related to #598.

And I add `jsComment` to `jsDestructuringBlock` and
`jsDestructuringArray`.


amadeus:
First off, thanks for taking the time and opening this PR!

However I think I'd like to rehaul how this is done to be more inline with how the rest of the file works. We tend to abuse the shit out of `nextgroup` and `skipwhite`, and `skipempty`. This means you'd need to probably create a few new specific contained `async` matches and then `nextgroup` them. I can work on this over the next few days if what I am saying doesn't make sense.

Arch-Mage:
Sorry If I can't understand what you're saying well (my english is not that good). By the way, what I can understand is that inserting optional `async` part into `jsFuncName` regex doesn't match the current project (maybe leading to spaghetti code :smile: ). I agree.
Any way, my Idea is to make `async` as an optional part of function declaration (it can only appear exactly before func declaration). That way, it won't messed up with `async` module that already popular. And now, the only way I can think of is to wrap function declaration into a single region. That's only my idea. I don't know if there will be any drawbacks, because I'm not that good at vim scripting.
Maybe this should be closed then?

PR #603, opened 2016-08-17
settings

bounceme:
these don't seem to be very relevant when using indentexprs, and may override the indentation script,

opinions on this are welcomed, and honestly I don't know exactly what effect these have. the clojure indent file and a few others included with vim has these options set to off


amadeus:
I am good with this, and if other indent plugins are also doing it, then I think it's a good idea for us to do it as well. The only thing I'd be worried about, is if for some reason, an older version of vim doesn't support those options, and could throw an error?

bounceme:
those have been in vim since 7.0 or earlier

PR #604, opened 2016-08-17
Develop -> Master 

bounceme:
All indent related. By the way, this will be making it's way into vim(as an update to the slightly outdated file from this repo already in the newest releases), so please try it and report bugs! 

I know of no regressions caused by these commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,11 @@ variables:
     let g:javascript_conceal_super          = "Ω"
     let g:javascript_conceal_arrow_function = "⇒"
 
+## Indentation Specific
+
+* `:h cino-:`
+* End-of-line continuations : `g:javascript_continuation` ( slightly complicated, look at the source code )
+* Start-of-line continuations : `g:javascript_opfirst` ( same as above )
 
 ## Contributing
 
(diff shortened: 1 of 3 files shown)


bounceme:
i'll probably squash this down when I merge (not going to rebase it though) since they mostly consist of iterative changes which over-wrote each other. With this update their shouldn't be any noticeable changes to the results of indenting,though the code is more readable and robust

bounceme:
@amadeus I think this can be merged, i'm prepared to fix any resulting bugs if i'm wrong about the stability. Doesn't seem anyone else is testing this either

Later on tonight or tomorrow?

amadeus:
I want to look into the `async` stuff before this gets merged. Also lets NOT squash this into master, I'd like to keep the history intact, in case we want to revert a specific commit or something, or need to `git bisect`.

I'll try to make a point of getting the `async` stuff started tonight. I also have a small flow bug or two that I'd like to get fixed as well.

bounceme:
have any changes for this update @amadeus ? if not, tomorrow i'll merge, otherwise i will wait if you care for making this a full repo update

bounceme:
better to get this out now is my opinion. It is a moderately sized indent update, and seems stable

PR #614, opened 2016-08-24
Fix export default style indents

amadeus:
/cc @bounceme 

I found an issue when using `export default` statements, indentation would not work properly.

I saw how you fixed an earlier import statement and modeled it after that.

Before this PR:
```js
export default {
    'key': 'val',
        'secondKey': 'val'
}
```

After this PR:
```js
export default {
    'key': 'val',
    'secondKey': 'val'
}
```


bounceme:
Thanks!

PR #621, opened 2016-09-02
adjustable indent (experimental)

bounceme:
https://github.com/mxw/vim-jsx/issues/116#issuecomment-244400990
thanks to @mxw for the suggestion, it is slightly less useful than the cindent version though

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,6 @@ variables:
 ## Indentation Specific
 
 * `:h cino-:`
-* End-of-line continuations : `g:javascript_continuation` ( deprecated )
-* Start-of-line continuations : `g:javascript_opfirst` ( deprecated )
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -100,8 +100,8 @@ variables:
 ## Indentation Specific
 
 * `:h cino-:`
-* End-of-line continuations : `g:javascript_continuation` ( slightly complicated, look at the source code )
-* Start-of-line continuations : `g:javascript_opfirst` ( same as above )
+* End-of-line continuations : `g:javascript_continuation` ( deprecated )
+* Start-of-line continuations : `g:javascript_opfirst` ( deprecated )
 
 ## Contributing
 
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -182,7 +182,7 @@ function GetJavascriptIndent()
   call cursor(b:js_cache[1],b:js_cache[2])
 
   let pline = getline(l:lnum) =~# s:expr_case ? getline(l:lnum) : substitute(getline(l:lnum), '\%(:\@<!\/\/.*\)$', '','')
-  let inb = num == 0 || ((l:line !~ s:line_pre . ',' && pline !~ ',' . s:line_term) || s:IsBlock())
+  let inb = num == 0 || s:IsBlock()
   let switch_offset = num == 0 || s:OneScope(num, strpart(getline(num),0,b:js_cache[2] - 1),1) !=# 'switch' ? 0 :
         \ &cino !~ ':' || !has('float') ?  s:sw() :
         \ float2nr(str2float(matchstr(&cino,'.*:\zs[-0-9.]*')) * (&cino =~# '.*:[^,]*s' ? s:sw() : 1))
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -122,7 +122,7 @@ endfunction
 
 function GetJavascriptIndent()
   if !exists('b:js_cache')
-    let b:js_cache = [0,0,0]
+    let b:js_cache = [0,0,0,0]
   endif
   " Get the current line.
   let l:line = getline(v:lnum)
@@ -152,9 +152,11 @@ function GetJavascriptIndent()
 
   " the containing paren, bracket, curly. Memoize, last lineNr either has the
   " same scope or starts a new one, unless if it closed a scope.
+  let known = 0
   call cursor(v:lnum,1)
   if b:js_cache[0] >= l:lnum  && b:js_cache[0] < v:lnum && b:js_cache[0] &&
         \ (b:js_cache[0] > l:lnum || s:Balanced(l:lnum) > 0)
+    let known = 1
     let num = b:js_cache[1]
   elseif syns != '' && l:line[0] =~ '\s'
     let pattern = syns =~? 'block' ? ['{','}'] : syns =~? 'jsparen' ? ['(',')'] :
@@ -163,7 +165,11 @@ function GetJavascriptIndent()
   else
     let num = s:GetPair('[({[]','[])}]','bW',2000)
   endif
-  let b:js_cache = [v:lnum,num,line('.') == v:lnum ? b:js_cache[2] : col('.')]
+
+  if known && b:js_cache[3]
+    let known = indent(b:js_cache[0]) - b:js_cache[3]
+  endif
+  let b:js_cache[:3] = [v:lnum,num,line('.') == v:lnum ? b:js_cache[2] : col('.')]
 
   if l:line =~ s:line_pre . '[])}]'
     return indent(num)
@@ -182,9 +188,13 @@ function GetJavascriptIndent()
   if inb && !swcase && ((l:line =~# g:javascript_opfirst || pline =~# g:javascript_continuation) ||
         \ num < l:lnum && s:OneScope(l:lnum,pline,0) =~# '\<\%(for\|each\|if\|let\|no\sb\|w\%(hile\|ith\)\)\>' &&
         \ l:line !~ s:line_pre . '{')
-    return (num > 0 ? indent(num) : -s:sw()) + (s:sw() * 2) + switch_offset
+    let ind = (num > 0 ? indent(num) : -s:sw()) + (s:sw() * 2) + switch_offset
+    let b:js_cache[3] = ind
+    return ind + known
   elseif num > 0
-    return indent(num) + s:sw() + switch_offset
+    let ind = indent(num) + s:sw() + switch_offset
+    let b:js_cache[3] = ind
+    return ind + known
   endif
 
 endfunction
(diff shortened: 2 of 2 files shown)


bounceme:
@sassanh please test this out, ive made the changes you wanted

bounceme:
here is a small demo
https://asciinema.org/a/3j66lltfis3p0mhhtgnycpz48

sassanh:
@bounceme thank you, that was fast. It's alright overall though there are some minor issues. Look at this:
```
const a = <div
    a={1 +
            2 +
            3 +
            4 +
    }
    b={true
            ? 1
            : 2
    }
    c={true
            ? 1
            : 2 && (true &&
                    false &&
                    false &&
                    true) && (true &&
                            false &&
                            false &&
                            true) && (
                                true &&
                                    false &&
                                    true
                            )
    }>123
</div>;
```
While rest of expressions are reasonable and can be considered they're following a rule, last parenthesis is inconsistent.

bounceme:
``` 
true) && (
                                true &&
                                    false &&
                                    true
                            )
```
that is because a continuation is started. the previous line ended in a parenthesis

mxw:
What's a "continuation"?  Is it a JS construct I'm not aware of?

bounceme:
not really, `:h cino-+`

mxw:
I agree with @sassanh that

```
var foo = (
    true &&
        false &&
        false
);
```

is not a "correct" indentation—I don't know of any style guidelines (save perhaps the most bizarre Python styles) where this is recommended.  I think I empathize more with the difficulties of trying to maintain a cobbled-together pseudo-parser—but I don't think the implementation details are a good justification for behavior.

mxw:
(I don't know the vim-javascript codebase at all, though, so for all I know this is prohibitively difficult to implement.)

bounceme:
https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L83 is the extent of this 'pseudo parser' . I won't be removing line continuation, though you can with the config variables https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L63

bounceme:
It could be changed to only happen in blocks(not in object literals, arrays, args), as with cindent

sassanh:
This is another issue I found with this branch:
```
                <X
                    itemKey='id'
                    template={props =>
                            <QuestionTag {...props.item} readonly={true}/>}
                            callback={(event, _, __, items) => }
                />
```
`callback` should be indented as same level as `template`.

bounceme:
are you sure that isn't an issue with jsx only?

sassanh:
It happened after I checked out this branch, I'm sure it was working correctly in `develop` and `jsx` branches.

bounceme:
as this is the javascript repo, could you please reduce this too a reproducible js only example?

bounceme:
https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244528553
That is something I actually will do

sassanh:
I just hope after you're done with it, it doesn't affect experience with `vim-jsx` if it doesn't improve it.

sassanh:
Another issue:
```
if (true &&
    false) {
        const a = 1;
    }
```
last line should have indent width 0.

bounceme:
thats unrelated, lines that start with `}])` are indented to the line which contains the other pair. wontfix

sassanh:
I see, you're right, it happens in `develop` branch as well.

bounceme:
https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244526767
is fixed

sassanh:
Very nice, this latest commit made it much better. I continue testing and will share any problem I find here.

sassanh:
A minor issue:
```
const a = true
    ? {
        zIndex: '1',
        backgroundColor: this.props.selected
        ? '#a0c4da'
        : '#cccccc',
        border: '1px solid #999999',
    }
    : {};
```
The ternary if in backgroundColor is not indented at all. Though I'm totally cool with its current indentation just thought it helps to report it.

bounceme:
that contradicts https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244526767

bounceme:
is that a github bug?^^

sassanh:
I see, it's alright the way it is.

bounceme:
I'll change it for objects if that is better

sassanh:
We can try it.

bounceme:
it makes no sense though, why not arrays? or parens?

sassanh:
Yeah, array or parens/args will work.

bounceme:
sure, https://github.com/pangloss/vim-javascript/blob/174d94e6f097851d495ba5cf0f8f537a9bcade62/indent/javascript.vim#L83 will work with those as well

mxw:
> It could be changed to only happen in blocks(not in object literals, arrays, args), as with cindent

Yeah, that sounds like the right constraint to me.  Thanks for your work, @bounceme!  (Also thanks for jumping in on the `vim-jsx` thread earlier :] )

bounceme:
https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244549552 I agree, and that's what we have now. thanks!

bounceme:
Sorry, I'm confused now. I have local changes that make this behaviour:
```
var foo = (
        true &&
        false &&
        false
);
foo(
        true &&
                false &&
                false
);
```
is that better or worse?

sassanh:
> is that better or worse?

I'd say it's definitly worse. Btw came here to thank you. It's just awesome now the way it is. There are some really minor issues that I even doubt are related to these commits. But overall it's just awesome.

bounceme:
it fixes https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244531668 though. commas inside of arrays,objects,args still don't create indent

sassanh:
https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244531668 is really minor imo comparing to https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244558129 which doesn't look good at all.

bounceme:
got it, it is the same idea, it just looks strange with that example

sassanh:
Currently your code snippet indents like this:
```
var foo = (
    true &&
    false &&
    false
);
foo(
    true &&
    false &&
    false
);
```
Which is exactly what I expect it. I which my C++ indenter would indent like this, I guess I should start a thread in their repo ð

bounceme:
I don't know, but i'll be sticking to https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244528553

bounceme:
do you think this is even necessary now ? https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L63

sassanh:
I don't know about that, I don't think so.

But look at this:
```
componentDidMount() {
        ::this.fetch();
}
```
double indent.

bounceme:
fixed

sassanh:
Didn't find any other issues with this branch. Checked it with ~2000 lines of js and jsx code.

sassanh:
A really minor issue is it somehow tries to indent commented lines too :D 
```
                {/* <div style={styles.y}> */}
                    {/*     <Z */}
                        {/*         {...this.props.x} */}
                        {/*         readonly={true} */}
                        {/*     /> */}
                    {/* </div> */}
                {/* <div style={styles.y}> */}
                    {/*     <Z */}
                        {/*         {...this.props.x} */}
                        {/*         readonly={true} */}
                        {/*         resource='question-label' */}
```
But I guess it's not relevant to these commits.

bounceme:
that is an issue with mxw/vim-jsx isn't it? i can't reproduce with ft=javascript

bounceme:
since they are so prone to user error, and they aren't so needed now I'll add a deprecation warning in the readme

sassanh:
I found an issue:
```
if (prevProps.x !==
    this.props.y) {
    this.setState({
        x: 1
    });
    }
```

bounceme:
https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244529942

sassanh:
Oh, right, my bad.

sassanh:
What about this:
```
if (true &&
    false) {
    a({
        1: 2,
    });
        b = 1;
    }
```
`b=1` has wrong indentation.

bounceme:
my results:
```
if (true &&
  false) {
    a({
      1: 2,
    });
    b = 1;
  }
```

sassanh:
Yeah I got latest `develop` version and it's fixed.

sassanh:
Are you using a variable to keep some state for indentation? After using it a while it will start indenting things like this https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244587987 Like indenter didn't clean a variable or something like that.

bounceme:
Yes, though it will never do things like that because it is used only in very specific circumstances. I did just fix something though which would cause some problems with the cache

sassanh:
OK, I will tell you if I see it again.

bounceme:
yea, only versions after https://github.com/pangloss/vim-javascript/commit/22d0e64ee15c906685cab631309e83882d692e38 had the problem

sassanh:
Nice, it's much more stable after your latest commits. I can't even find minor issues. Good job.

bounceme:
If there were any jsx bugs found here, you might report them at mxw/vim-jsx so they get recognized

sassanh:
It happened again, in a vim session that was open for few hours, indentation went wrong. After restarting vim it worked alright.

bounceme:
so what do i do with that information? the conditions absolutely prevent that from happening, my guess is jsx

sassanh:
Yeah, I know that information is not helpful. Just though maybe it triggers something for you. I try to find out how to reproduce it.

bounceme:
Well I have noticed jsx changing indent when repeatedly indenting

sassanh:
Next time it happens I'll check a js file with `ft=javascript`. Then we can say for certain if it's related to `vim-javascript` at all or not. Last time it happened I was checking 3rd line of a js file (but with `ft=javascript.jsx`) the line it indented incorrectly wasn't in any JSXRegion. It was as simple as this:
```
import {x} from './y';
    import z from './w';
```

PR #624, opened 2016-09-05
flow autodetect

bounceme:
this is much simpler than autocmds

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -213,7 +213,7 @@ if exists("javascript_plugin_jsdoc")
   endif
 endif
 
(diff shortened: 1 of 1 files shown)


bounceme:
this also mostly removes the need for a variable, however new files won't automatically have flow highlighting till they are saved, or the buffer gets re-entered

bounceme:
also would prevent always loading this syntax for every plain js file

amadeus:
Sorry I've been slow to respond lately, I've been out of town. The variable is still a requirement because you can set your entire project to flow, which won't require the `@flow` detection.

PR #625, opened 2016-09-05
jsx indent

bounceme:
Just experimenting, @mxw , @sassanh any opinions?

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -180,9 +180,7 @@ function GetJavascriptIndent()
 
   let b:js_cache[:2] = [v:lnum,num,line('.') == v:lnum ? b:js_cache[2] : col('.')]
 
-  if s:xml && search('^\s*\S\%' . v:lnum .'l') &&
-        \ (l:line =~ '^\s*\%(\/>\|<\/\)' || search('\%'.v:lnum.'l^\s*<\w*\_s*\/>') ||
-        \ s:GetPair('<[^? \t>/]\+\%(\_s*>\|\_s\+\_[^>]*[^/]>\)','</[^? \t>/]\+>','cbnW',200))
+  if s:xml && syns =~? 'xml\|jsx'
     return XmlIndentGet(v:lnum, 0)
   endif
 
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -8,6 +8,10 @@
 if exists('b:did_indent')
   finish
 endif
+if !exists('s:xml')
+  runtime! indent/xml.vim
+  let s:xml = exists('*XmlIndentGet')
+endif
 let b:did_indent = 1
 
 " Now, set up our indentation expression and keys that trigger it.
@@ -175,6 +179,12 @@ function GetJavascriptIndent()
   endif
 
   let b:js_cache[:2] = [v:lnum,num,line('.') == v:lnum ? b:js_cache[2] : col('.')]
+
+  if s:xml && getline(b:js_cache[1])[b:js_cache[2]-1] == '(' && search('^\s*\S\%' . v:lnum .'l') &&
+        \ (l:line =~ '^\s*\%(\/>\|<\/\)' || search('\%'.v:lnum.'l^\s*<\w*\_s*\/>') ||
+        \ s:GetPair('<[^? \t>/]\+\%(\_s*>\|\_s\+[^>]*[^/]>\)','</[^? \t>/]\+>','cbnW',200))
+    return XmlIndentGet(v:lnum, 0)
+  endif
 
   if l:line =~ s:line_pre . '[])}]'
     return indent(num)
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -180,9 +180,9 @@ function GetJavascriptIndent()
 
   let b:js_cache[:2] = [v:lnum,num,line('.') == v:lnum ? b:js_cache[2] : col('.')]
 
-  if s:xml && getline(b:js_cache[1])[b:js_cache[2]-1] == '(' && search('^\s*\S\%' . v:lnum .'l') &&
+  if s:xml && search('^\s*\S\%' . v:lnum .'l') &&
         \ (l:line =~ '^\s*\%(\/>\|<\/\)' || search('\%'.v:lnum.'l^\s*<\w*\_s*\/>') ||
-        \ s:GetPair('<[^? \t>/]\+\%(\_s*>\|\_s\+[^>]*[^/]>\)','</[^? \t>/]\+>','cbnW',200))
+        \ s:GetPair('<[^? \t>/]\+\%(\_s*>\|\_s\+\_[^>]*[^/]>\)','</[^? \t>/]\+>','cbnW',200))
     return XmlIndentGet(v:lnum, 0)
   endif
 


sassanh:
Once I suggested using html instead of xml to @mxw and he had his reasons to stick with xml. But locally I have modified `vim-jsx` to use html instead of xml. Though I agree with @mxw about his reasons, I think using html has lots of benefits. Lots of plugins written for html work out of the box for jsx. I have absolutely no problem with my modified version. Everything works alright. So I suggest you to consider html instead of xml too.

You can take a look at it here: https://github.com/sassanh/vim-jsx
It needs this modified html syntax in your ~/.vim/syntax/html.vim: https://gist.github.com/sassanh/ed6e07fecde5f5c063d1af3852c7eb5f

Let me know if there's any problem with it, maybe it relies on other files in my setup too.

sassanh:
Btw, in my neovim setup with `xml.vim` and `editxml` installed `exists('*XmlIndentGet')` returns `0`. So I can't test it.

sassanh:
Ok I had to `:runtime indet/xml.vim`.

indent/javascript.vim
```diff
@@ -176,6 +180,12 @@ function GetJavascriptIndent()
 
   let b:js_cache[:2] = [v:lnum,num,line('.') == v:lnum ? b:js_cache[2] : col('.')]
 
+  if s:xml && getline(b:js_cache[1])[b:js_cache[2]-1] == '(' && search('^\s*\S\%' . v:lnum .'l') &&
```
sassanh: While running XmlIndentGet returns reasonable indentation for the line, I wasn't able to write a jsx code that satisfies conditions in this if. So it applies normal javascript indentation.

sassanh:
I can't satisfy this if: https://github.com/pangloss/vim-javascript/pull/625/files#diff-564c1f6e79518959361f7ad8a86f5792R183

bounceme:
You'll need to show some valid example code

bounceme:
also disable the jsx plugin for testing

sassanh:
`vim-jsx` was disabled. A simple sample:
```
return <div
ref='container'
```
second line should be indented.

bounceme:
Is that actually valid ? or are you leaving parts out? I thought that all html/xml needed to be in one line unless if it was in parentheses

bounceme:
anyway, the reason i will use the xml script and the reason I won't be covering edge cases like that is that this will be a minimal level of support if it gets added

sassanh:
It is valid. This is valid. Let me give you a sample to work on:
```
export default class Barcode extends React.PureComponent {
    componentDidMount() {
        this.props.callback(this.refs.container.outerHTML);
    }

    render() {
        return <div
        ref='container'
            style={{
                width: '148mm',
                height: '210mm',
                position: 'relative',
            }}>
            <Box
                top={barcodeTop}
                left={barcodeLeft}
            />
            <Box
                top={barcodeTop}
                left={barcodeLeft + barcodeCell * 31}
            />
            <Box
                top={barcodeTop + barcodeCell * 2}
                left={barcodeLeft}
            />
            <Box
                top={barcodeTop + barcodeCell * 2}
                left={barcodeLeft + barcodeCell * 31}
            />
            {_.map(Array(90), (_i, index) => {
                const x = index % 30 + 1;
                const y = Math.floor(index / 30);
                const i = index % 45;
                return <Box
                    key={index}
                    top={barcodeTop + y * barcodeCell}
                    left={barcodeLeft + x * barcodeCell}
                />;
            })}
            <div
                style={{
                    position: 'absolute',
                    top: '50mm',
                    right: '40mm',
                }}
                >
                {this.props.studentNumber}
            </div>
            <div
                style={{
                    position: 'absolute',
                    top: '58mm',
                    right: '40mm',
                }}
                >
                {this.props.firstname}
            </div>
        </div>;
    }
}
```

sassanh:
> anyway, the reason i will use the xml script and the reason I won't be covering edge cases like that is that this will be a minimal level of support if it gets added

Then please add an option to disable it. I love the idea of supporting jsx in `vim-javascript` but currently my personal setup is really mature and it'd be bothersome if `vim-javascript` breaks it.

bounceme:
Is that really plain jsx though? you aren't using any babel compilation or anything else

sassanh:
I'm using babel transforms for es7 javascript. But the jsx parts are just plain jsx.

sassanh:
Look at this: https://jsfiddle.net/reactjs/n47gckhr/?utm_source=website&utm_medium=embed&utm_campaign=n47gckhr
It's from here: https://facebook.github.io/react/docs/thinking-in-react.html

bounceme:
well, if this isnt the case: https://github.com/pangloss/vim-javascript/pulls#issuecomment-244807875
then this just won't happen

bounceme:
actually, try removing this: `getline(b:js_cache[1])[b:js_cache[2]-1] == '(' &&`

mxw:
Is this diff trying to implement JSX indentation as part of vim-javascript?  If so, are you planning to support syntax also?  Indentation without syntax highlighting seems counterintuitive, and I suspect that this will break vim-jsx's indentation (which will require users to either use only vim-javascript, or to use vim-jsx with some other JS package).

I prefer modularity in vim packages (and software more generally), so while I won't make too much of a fuss if this is really the direction you want to go in, I don't see what the benefits are over having vim-javascript as a standalone package (and I _do_ see downsides).

The reason vim-jsx uses XML instead of HTML as the dependency is that HTML is incorrect.  JSX simply is not "JavaScript with embedded HTML"; it's "JavaScript with embedded XML".

bounceme:
I agree, i was really just curious since i thought that the indent here was able enoguh to make it fairly easy

mxw:
Is there still some problem that you're trying to solve here, after fixing the line continuation behavior to match cindent?

vim-jsx is a very basic package, which is another reason I think it's better for it to be independent—it makes debugging it a lot easier, and it's pretty trivial for users to include or not include it based on whether they care about JSX, and its surface area of interaction with the dependency is minimal.  It's not like vim does much in the way of supporting interpackage APIs, so things that are much more complex than vim-jsx probably need to be integrated.

bounceme:
One simple way of maintaining compatibility would be to just `finish` if the `b:js_cache` variable exists. that's recent enough, or i could make the [`s:xml`](https://github.com/pangloss/vim-javascript/pull/625/files#diff-564c1f6e79518959361f7ad8a86f5792R13) var buffer local

bounceme:
you make  a good point however as those solutions aren't that elegant

bounceme:
I'll probably just see if it is improve-able than save whatever I end up with in a gist, or market it to the e4x using hipsters

bounceme:
hey, this is actually working very well in my testing can some others voice their opinions after testing?

mxw:
What bugs in vim-jsx is it fixing?

bounceme:
None that i know of, i don't even write jsx, it is a small extension to this indent script though

mxw:
Is it compatible with vim-jsx's indenter?  If not, what is your intended solution for folks who want to use vim-javascript but want both JSX indentation and highlighting?

More broadly, I'm not sure why this is desirable.  If this is no better than an independent module for handling the JSX extension to JS, then to me, it seems strictly worse, since it's tightly coupling vim support for two things which have a dependent rather than a coupled relationship.  If it _were_ better, that'd be a different story for sure.

mxw:
(Although I would argue that in that case, the solution would be for me or someone else to make `vim-jsx` suck less before resorting to coupling support for JSX into `vim-javascript`.)

bounceme:
Well if you compare the code added here to the indent script in vim-jsx you can see some benefits in terms of size, the only reason i'm trying this is because of how easy to fit in this plugin it is

bounceme:
so, I think there is almost no reason, in terms of complexity or speed, for this addition ( if it really is working ). it relies on mxw/jsx .

bounceme:
and i'm going to assume this wouldn't work in the jsx plugin (though I haven't tried)

mxw:
If it relies on `vim-jsx`, did you verify that it's being executed?

`vim-jsx`'s indentation logic straightforward—it decides whether we're in XML or in JS, and invokes the appropriate indentation routine, making some adjustments at the boundaries.  Under what conditions is your codepath being hit?

bounceme:
i deleted the jsx indent file

mxw:
Certainly the current version of this PR depends on JSX.  If it's a no-op, then, well, it's not needed.  If it _isn't_ a no-op, then I'm confused about how it could be correct—unless it's fixing a known issue with `vim-jsx`, if this diff does anything, then `vim-jsx` will "correct" for it, and subsequently be wrong.

mxw:
I'm confused; this is what I'm seeing:

screen shot 2016-09-05 at 11 02 34 pm

bounceme:
yea, thats all. it is reliant on the jsx syntax file only

mxw:
But why would you want JSX indentation support inside `vim-javascript` without syntax support?  Having a two-way dependency with `vim-jsx` seems... convoluted.

mxw:
By the by, this is a cobbled-together file that I use for sanity testing `vim-jsx`: https://gist.github.com/mxw/ad21ea4194fcf19d3e12f0ac77f2edac

If your indenter is a no-op on it, then it isn't obviously wrong :]

bounceme:
Yea, this was my test file (and results)
```
// Awkward block of variable declarations
var Form = MyFormComponent;
var FormRow = Form.Row;
var FormLabel = Form.Label;
var FormInput = Form.Input;

var App = (
  <Form>
    <FormRow>
      <FormLabel />
      <FormInput />
    </FormRow>
  </Form>
);
export default class Barcode extends React.PureComponent {
  componentDidMount() {
    this.props.callback(this.refs.container.outerHTML);
  }

  render() {
    return <div
      ref='container'
      style={{
	width: '148mm',
	height: '210mm',
	position: 'relative',
      }}>
      <Box
	top={barcodeTop}
	left={barcodeLeft}
	/>
      <Box
	top={barcodeTop}
	left={barcodeLeft + barcodeCell * 31}
	/>
      <Box
	top={barcodeTop + barcodeCell * 2}
	left={barcodeLeft}
	/>
      <Box
	top={barcodeTop + barcodeCell * 2}
	left={barcodeLeft + barcodeCell * 31}
	/>
      {_.map(Array(90), (_i, index) => {
	const x = index % 30 + 1;
	const y = Math.floor(index / 30);
	const i = index % 45;
	return <Box
	key={index}
	top={barcodeTop + y * barcodeCell}
	left={barcodeLeft + x * barcodeCell}
	/>;
      })}
      <div
	style={{
	  position: 'absolute',
	  top: '50mm',
	  right: '40mm',
	}}
	>
	{this.props.studentNumber}
      </div>
      <div
	style={{
	  position: 'absolute',
	  top: '58mm',
	  right: '40mm',
	}}
	>
	{this.props.firstname}
      </div>
    </div>;
  }
}
```
I think there is condition with syn names that this would work,it is the js being highlighted as jsxregion that messes up

mxw:
That doesn't look right...

I'm guessing that you aren't handling interleaved JS and JSX regions correctly.  Coincidentally, I know of a few lines of Vimscript created for specifically that purpose...

bounceme:
sorry, was testing with the default syntax, not this repos

bounceme:
results with the pangloss syntax
```
// Awkward block of variable declarations
var Form = MyFormComponent;
var FormRow = Form.Row;
var FormLabel = Form.Label;
var FormInput = Form.Input;

var App = (
  <Form>
    <FormRow>
      <FormLabel />
      <FormInput />
    </FormRow>
  </Form>
);
export default class Barcode extends React.PureComponent {
  componentDidMount() {
    this.props.callback(this.refs.container.outerHTML);
  }

  render() {
    return <div
      ref='container'
      style={{
	width: '148mm',
	height: '210mm',
	position: 'relative',
      }}>
      <Box
	top={barcodeTop}
	left={barcodeLeft}
	/>
      <Box
	top={barcodeTop}
	left={barcodeLeft + barcodeCell * 31}
	/>
      <Box
	top={barcodeTop + barcodeCell * 2}
	left={barcodeLeft}
	/>
      <Box
	top={barcodeTop + barcodeCell * 2}
	left={barcodeLeft + barcodeCell * 31}
	/>
      {_.map(Array(90), (_i, index) => {
	const x = index % 30 + 1;
	const y = Math.floor(index / 30);
	const i = index % 45;
	return <Box
	  key={index}
	  top={barcodeTop + y * barcodeCell}
	  left={barcodeLeft + x * barcodeCell}
	  />;
      })}
      <div
	style={{
	  position: 'absolute',
	  top: '50mm',
	  right: '40mm',
	}}
	>
	{this.props.studentNumber}
      </div>
      <div
	style={{
	  position: 'absolute',
	  top: '58mm',
	  right: '40mm',
	}}
	>
	{this.props.firstname}
      </div>
    </div>;
  }
}
```

mxw:
That still seems pretty off.  Again, I'm just not sure what problem we're trying to solve here.  I'm pretty sure the startup cost of loading an extra indent script is not that bad.

bounceme:
I have no idea why that looks different on github than in vim

bounceme:
oh, ive got some weird tabs + spaces in this file

bounceme:
It's just that i'd rather have 3 lines of code than ~100 if there isn't a difference in result. This can be used in your repo if you want, like it says on the readme pangloss/vim-javascript has the best results anyway

sassanh:
While I'm following your discussion and am curious to see how it goes. Just to give feedback, https://github.com/pangloss/vim-javascript/pull/625#issuecomment-244844651 seems much better than your first commit. Could you please push your latest commits?
```
<Box
  top={barcodeTop}
  left={barcodeLeft}
  />
```
last line should have same indentation as first line. like this:
```
<Box
  top={barcodeTop}
  left={barcodeLeft}
/>
```
while it doesn't apply for this:
```
<div
  x="y"
  >
</div>
```

sassanh:
The gist file Max referenced is great jsx sample. You should indent it exactly the way it is.

bounceme:
they are pushed, at least i think they are? also changing xml.vim 's indent result can be done as well if it matters to have the aligned `/>` . just proof of concept at the moment, which seems to be working fine, though not error proof yet

bounceme:
my results @sassanh  
```
static propTypes = {
  numberInGroup: PropTypes.number,
}

import React from 'react'
import { Row, Col } from 'react-flexbox-grid'
import AppBar from 'material-ui/lib/app-bar'
import IconButton from 'material-ui/lib/icon-button'
import NavigationArrowBack from 'material-ui/lib/svg-icons/navigation/arrow-back'
import LoginForm from './components/LoginForm'

class Test extends Component {
  render() {
    return (
      <div>
        {
          Object.keys(test).map(
            return (
              <bar>
              </bar>
            );
          )
        }
      </div>
    );
  }
}

export default React.createClass({
  contextTypes: {
    router: React.PropTypes.object.isRequired,
  },

  render() {
    return (
      <div>
        banana hyphone
        <AppBar title="CatCoin" iconElementLeft={ <IconButton onTouchTap={this.back}><NavigationArrowBack /></IconButton> } />
        <Row center="xs">
          <Col xs={8} sm={8} md={8} lg={6}>
            <LoginForm onLogin={ ()=> {alert(0)}} />
          </Col>
        </Row>
      </div>
    )
  },

  back() {
    this.context.router.push('/')
  },

  login(username, password) {
    alert(0)
  },
})

function SearchTags3 (props) {
  return (
    <ul className='SearchTags'>
      {props.searches.map((text) => {
        return (
          <li key={text}>
            {text}
            <button onMouseDown={preventDefault(() => props.removeSearchTag(text))}>
              &times;
            </button>
          </li>
        )
      })}
    </ul>
  )
}

function renderEmpty() {
  return (
    <div className='grid'>
      <div></div>
    </div>
  );
}

// Input (JSX):
var person = <Person name={window.isLoggedIn ? window.name : ''} />;

// Input (JSX):
var content = <Container>{window.isLoggedIn ? <Nav /> : <Login />}</Container>;

  <div onClick={() => console.log('foo')} className="will not be highlighted" />

  /*
   * NB: Highlighting for the comments within the <Person> node is known to be
   * broken.
   */
  var content = (
    <Nav>
      {/* child comment, put {} around */}
      <Person
        /* multi
        line
        comment */
        name={window.isLoggedIn ? window.name : ''} // end of line comment
        />
    </Nav>
  );

var Form = MyFormComponent;
var FormRow = Form.Row;
var FormLabel = Form.Label;
var FormInput = Form.Input;

var App = (
  <Form>
    <FormRow>
      <FormLabel />
      {"banana"}
      <FormInput />
    </FormRow>
  </Form>
);

<div>
  <Nav />
  <Button />
  <Table />
  {/*
    <Modal />
    */
  }
</div>;

export default ({foo, bar}) => (
  <div className="sample" key={foo}>
    <div
      id="nested sample"
      sample={{
        a: 1,
        b: 'sample',
      }}
      />
  </div>
);

  <Navbar.Form pullRight>
    {
      auth.loggedIn()
      ? <Button onClick={this.logout}>Log Out</Button>
      : <Button bsStyle="primary" onClick={auth.login}>Log In</Button>
    }
  </Navbar.Form>

  var blah =
  <div>
    { loading ?
      <foo /> :
      <bar>
        <baz>
          "blah"
        </baz>
      </bar>
    }
  </div>;
```

sassanh:
I doubt if `jsx` branch has your latest code, it indents like this:
```
const Box = props => <div
style={{
```

bounceme:
delete the indent file in the jsx repo, set ft to javascript.jsx

mxw:
@bounceme—Oh yeah, that's true; what you have now has no dependency on `vim-javascript`.  I was mixing up what you have before with what you have now.

That said, what you currently have almost certainly is not correct in general.  There's a reason that `vim-jsx` does the stuff it does, and that reason is not that I have a particular love of verbosity—it's that some of the synstack inspection is necessary for correctness (and I designed the syntax elements so that the indentation would cooperate with it).

As the author of a package that depends on `vim-javascript`, I have a strong preference for you not landing this change.  If you do, you'll introduce a cross-dependency—for instance, if I change the way that `vim-jsx`'s syntax elements work or are named, I'll break `vim-javascript`, and thereby also break `vim-jsx`, the sole purpose of which is to glue a JS syntax/indent package to an XML syntax/indent package.

The cross-dependency is a huge nuisance as a dependent of `vim-javascript, so if you must land this, please at least consider adding an opt-out config for it.

bounceme:
of course, any change to make usability worse for vim-jsx users isn't deliberate. It is really just an experiment, I won't be merging this soon if at all

mxw:
Ah, okay—carry on then, as you please.

I do suspect that if you work through all the edge cases that make this different from what `vim-jsx` is doing, you'll end up with something a lot like `vim-jsx`.

PR #629, opened 2016-09-11
speed increase attempt

bounceme:


bounceme:
Testing with https://github.com/sanctuary-js/sanctuary-def/blob/master/index.js

before: ~11s
after: 3.86s

amadeus:
This is pretty huge if it doesn't introduce any side effects!

bounceme:
Definitely happy about this! also it is 100% accurate ( or could be if there is something i've missed ). It is only used for https://github.com/pangloss/vim-javascript/blob/21c3f0d2417b190ac4892d7cbbef64711b4a77f0/indent/javascript.vim#L175-L177 since that is the major bottleneck

PR #631, opened 2016-09-13
Fix Various Syntax Highlighting Issues

jfelchner:
I've annotated each commit with what/why of the changes.

This fixes 3 or 4 current outstanding issues.

I've also made some opinionated tweaks to the current highlight linkages.  Hopefully you all agree with them.  The improvements to the readability of the code I'm currently working on are substantial.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -326,8 +326,8 @@ if version >= 508 || !exists("did_javascript_syn_inits")
   HiLink jsModuleKeywords       Include
   HiLink jsModuleOperators      Include
   HiLink jsModuleDefault        Include
-  HiLink jsDecorator            Special
-  HiLink jsDecoratorFunction    Special
+  HiLink jsDecorator            Tag
+  HiLink jsDecoratorFunction    Tag
   HiLink jsFuncArgOperator      jsFuncArgs
   HiLink jsModuleAsterisk       Noise
   HiLink jsClassProperty        jsObjectKey
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -41,12 +41,12 @@ syntax keyword jsModuleKeywords  contained import
 syntax keyword jsModuleKeywords  contained export skipwhite skipempty nextgroup=jsExportBlock,jsModuleDefault
 syntax keyword jsModuleOperators contained from
 syntax keyword jsModuleOperators contained as
-syntax region  jsModuleGroup     contained matchgroup=jsBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
+syntax region  jsModuleGroup     contained matchgroup=jsModuleBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
 syntax match   jsModuleAsterisk  contained /*/
 syntax keyword jsModuleDefault   contained default skipwhite skipempty nextgroup=@jsExpression
 syntax region  jsImportContainer start=/\<import\> / end="\%(;\|$\)" contains=jsModuleKeywords,jsModuleOperators,jsComment,jsString,jsTemplateString,jsNoise,jsModuleGroup,jsModuleAsterisk
 syntax region  jsExportContainer start=/\<export\> / end="\%(;\|$\)" contains=jsModuleKeywords,jsModuleOperators,jsStorageClass,jsModuleDefault,@jsExpression
-syntax region  jsExportBlock     contained matchgroup=jsBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
+syntax region  jsExportBlock     contained matchgroup=jsExportBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
 
 " Strings, Templates, Numbers
 syntax region  jsString           start=+"+  skip=+\\\("\|$\)+  end=+"\|$+  contains=jsSpecial,@Spell extend
@@ -98,17 +98,17 @@ exe 'syntax keyword jsSuper     super  contained '.(exists('g:javascript_conceal
 " Statement Keywords
 syntax keyword jsStatement    contained break continue with yield debugger
 syntax keyword jsConditional            if           skipwhite skipempty nextgroup=jsParenIfElse
-syntax keyword jsConditional            else         skipwhite skipempty nextgroup=jsCommentMisc,jsBlock
+syntax keyword jsConditional            else         skipwhite skipempty nextgroup=jsCommentMisc,jsIfElseBlock
 syntax keyword jsConditional            switch       skipwhite skipempty nextgroup=jsParenSwitch
 syntax keyword jsRepeat                 while for    skipwhite skipempty nextgroup=jsParenRepeat
-syntax keyword jsDo                     do           skipwhite skipempty nextgroup=jsBlock
+syntax keyword jsDo                     do           skipwhite skipempty nextgroup=jsRepeatBlock
 syntax keyword jsLabel        contained case default
 syntax keyword jsTry                    try          skipwhite skipempty nextgroup=jsTryCatchBlock
-syntax keyword jsFinally      contained finally      skipwhite skipempty nextgroup=jsBlock
+syntax keyword jsFinally      contained finally      skipwhite skipempty nextgroup=jsFinallyBlock
 syntax keyword jsCatch        contained catch        skipwhite skipempty nextgroup=jsParenCatch
 syntax keyword jsException              throw
 syntax keyword jsAsyncKeyword           async await
-syntax match   jsSwitchColon  contained /:/          skipwhite skipempty nextgroup=jsBlock
+syntax match   jsSwitchColon  contained /:/          skipwhite skipempty nextgroup=jsSwitchBlock
 
 " Keywords
 syntax keyword jsGlobalObjects  Array Boolean Date Function Iterator Number Object Symbol Map WeakMap Set RegExp String Proxy Promise Buffer ParallelArray ArrayBuffer DataView Float32Array Float64Array Int16Array Int32Array Int8Array Uint16Array Uint32Array Uint8Array Uint8ClampedArray JSON Math console document window Intl Collator DateTimeFormat NumberFormat
@@ -136,16 +136,19 @@ syntax keyword jsHtmlEvents     onblur onclick oncontextmenu ondblclick onfocus 
 "" Code blocks
 syntax region  jsBracket                      matchgroup=jsBrackets            start=/\[/ end=/\]/ contains=@jsExpression extend fold
 syntax region  jsParen                        matchgroup=jsParens              start=/(/  end=/)/  contains=@jsAll extend fold
-syntax region  jsParenIfElse        contained matchgroup=jsParensIfElse        start=/(/  end=/)/  contains=@jsAll skipwhite skipempty nextgroup=jsCommentMisc,jsBlock extend fold
-syntax region  jsParenRepeat        contained matchgroup=jsParensRepeat        start=/(/  end=/)/  contains=@jsAll skipwhite skipempty nextgroup=jsCommentMisc,jsBlock extend fold
+syntax region  jsParenIfElse        contained matchgroup=jsParensIfElse        start=/(/  end=/)/  contains=@jsAll skipwhite skipempty nextgroup=jsCommentMisc,jsIfElseBlock extend fold
+syntax region  jsParenRepeat        contained matchgroup=jsParensRepeat        start=/(/  end=/)/  contains=@jsAll skipwhite skipempty nextgroup=jsCommentMisc,jsRepeatBlock extend fold
 syntax region  jsParenSwitch        contained matchgroup=jsParensSwitch        start=/(/  end=/)/  contains=@jsAll skipwhite skipempty nextgroup=jsSwitchBlock extend fold
 syntax region  jsParenCatch         contained matchgroup=jsParensCatch         start=/(/  end=/)/  skipwhite skipempty nextgroup=jsTryCatchBlock extend fold
 syntax region  jsFuncArgs           contained matchgroup=jsFuncParens          start=/(/  end=/)/  contains=jsFuncArgCommas,jsComment,jsFuncArgExpression,jsDestructuringBlock,jsRestExpression,jsFlowArgumentDef skipwhite skipempty nextgroup=jsCommentFunction,jsFuncBlock,jsFlowReturn extend fold
 syntax region  jsClassBlock         contained matchgroup=jsClassBraces         start=/{/  end=/}/  contains=jsClassFuncName,jsClassMethodType,jsArrowFunction,jsArrowFuncArgs,jsComment,jsGenerator,jsDecorator,jsClassProperty,jsClassPropertyComputed,jsClassStringKey,jsNoise extend fold
 syntax region  jsFuncBlock          contained matchgroup=jsFuncBraces          start=/{/  end=/}/  contains=@jsAll extend fold
+syntax region  jsIfElseBlock        contained matchgroup=jsIfElseBraces        start=/{/  end=/}/  contains=@jsAll extend fold
 syntax region  jsBlock              contained matchgroup=jsBraces              start=/{/  end=/}/  contains=@jsAll extend fold
-syntax region  jsTryCatchBlock      contained matchgroup=jsBraces              start=/{/  end=/}/  contains=@jsAll skipwhite skipempty nextgroup=jsCatch,jsFinally extend fold
-syntax region  jsSwitchBlock        contained matchgroup=jsBraces              start=/{/  end=/}/  contains=@jsAll,jsLabel,jsSwitchColon extend fold
+syntax region  jsTryCatchBlock      contained matchgroup=jsTryCatchBraces      start=/{/  end=/}/  contains=@jsAll skipwhite skipempty nextgroup=jsCatch,jsFinally extend fold
+syntax region  jsFinallyBlock       contained matchgroup=jsFinallyBraces       start=/{/  end=/}/  contains=@jsAll extend fold
+syntax region  jsSwitchBlock        contained matchgroup=jsSwitchBraces        start=/{/  end=/}/  contains=@jsAll,jsLabel,jsSwitchColon extend fold
+syntax region  jsRepeatBlock        contained matchgroup=jsRepeatBraces        start=/{/  end=/}/  contains=@jsAll extend fold
 syntax region  jsDestructuringBlock contained matchgroup=jsDestructuringBraces start=/{/  end=/}/  contains=jsDestructuringProperty,jsDestructuringAssignment,jsDestructuringNoise,jsDestructuringPropertyComputed,jsSpreadExpression extend fold
 syntax region  jsDestructuringArray contained matchgroup=jsDestructuringBraces start=/\[/ end=/\]/ contains=jsDestructuringPropertyValue,jsNoise,jsDestructuringProperty,jsSpreadExpression extend fold
 syntax region  jsObject                       matchgroup=jsObjectBraces        start=/{/  end=/}/  contains=jsObjectKey,jsObjectKeyString,jsObjectKeyComputed,jsObjectSeparator,jsObjectFuncName,jsObjectMethodType,jsGenerator,jsComment,jsObjectStringKey,jsSpreadExpression extend fold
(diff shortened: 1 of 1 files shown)


syntax/javascript.vim
```diff
@@ -25,6 +25,7 @@ syntax sync fromstart
 syntax case match
 
 syntax match   jsNoise          /[:,\;\.]\{1}/
+syntax match   jsSemiEndOfLine  /\;\{1}\(\(\s*\/[\/\*].*\)\?$\)\@=/
```
jfelchner: FWIW, this also handles the use case of `;` followed by comments

amadeus:
Thanks for opening this! It's gonna take me a it to go over this and test it!

syntax/javascript.vim
amadeus: Just at a quick glance, I don't like this one at all because it's not contained AND has a look ahead. This will probably destroy performance in a larger file. Also, I am not quite sure the rationale to match end of line semis? Wouldn't it be better to isolate particular semis in contained groups we already have?

syntax/javascript.vim
```diff
@@ -91,32 +94,33 @@ exe 'syntax keyword jsReturn    return contained '.(exists('g:javascript_conceal
...
-syntax keyword jsGlobalObjects  Array Boolean Date Function Iterator Number Object Symbol Map WeakMap Set RegExp String Proxy Promise Buffer ParallelArray ArrayBuffer DataView Float32Array Float64Array Int16Array Int32Array Int8Array Uint16Array Uint32Array Uint8Array Uint8ClampedArray JSON Math console document window Intl Collator DateTimeFormat NumberFormat
-syntax keyword jsGlobalNodeObjects module exports global process
-syntax match   jsGlobalNodeObjects /require/ contains=jsFuncCall
-syntax keyword jsExceptions     Error EvalError InternalError RangeError ReferenceError StopIteration SyntaxError TypeError URIError
-syntax keyword jsBuiltins       decodeURI decodeURIComponent encodeURI encodeURIComponent eval isFinite isNaN parseFloat parseInt uneval
+syntax keyword jsGlobalObjects      Array Boolean Date Function Iterator Number Object Symbol Map WeakMap Set RegExp String Proxy Promise Buffer ParallelArray ArrayBuffer DataView Float32Array Float64Array Int16Array Int32Array Int8Array Uint16Array Uint32Array Uint8Array Uint8ClampedArray JSON Math console document window Intl Collator DateTimeFormat NumberFormat
+syntax match   jsConstant           "\<[A-Z][A-Za-z]\+\>"
```
amadeus: We've had many some similar discussions in the past regarding things how to handle things constants and keywords, and whether to highlight them or not.

My personal opinion on the matter is something like this should NOT be in our syntax file because it's more of a stylistic convention (using capital letters to denote a constants), not a rule of the language.

You could easily add something like this as an `after` plugin. However just because you use an initial capital letter does not a constant make.

Also, this regex could be improved if you use it in an `after` plugin, since it will break if there are numbers or `$` in the name.

syntax/javascript.vim
```diff
@@ -217,8 +228,8 @@ if exists("javascript_plugin_flow")
   runtime extras/flow.vim
 endif
 
-syntax cluster jsExpression  contains=jsBracket,jsParen,jsObject,jsBlock,jsTernaryIf,jsTaggedTemplate,jsTemplateString,jsString,jsRegexpString,jsNumber,jsFloat,jsOperator,jsBooleanTrue,jsBooleanFalse,jsNull,jsFunction,jsArrowFunction,jsGlobalObjects,jsExceptions,jsFutureKeys,jsDomErrNo,jsDomNodeConsts,jsHtmlEvents,jsFuncCall,jsUndefined,jsNan,jsPrototype,jsBuiltins,jsNoise,jsClassDefinition,jsArrowFunction,jsArrowFuncArgs,jsParensError,jsComment,jsArguments,jsThis,jsSuper,jsDo
-syntax cluster jsAll         contains=@jsExpression,jsExportContainer,jsImportContainer,jsStorageClass,jsConditional,jsRepeat,jsReturn,jsStatement,jsException,jsTry,jsAsyncKeyword
+syntax cluster jsExpression  contains=jsConstant,jsBracket,jsParen,jsObject,jsBlock,jsTernaryIf,jsTaggedTemplate,jsTemplateString,jsString,jsRegexpString,jsNumber,jsFloat,jsOperator,jsBooleanTrue,jsBooleanFalse,jsNull,jsFunction,jsArrowFunction,jsGlobalObjects,jsExceptions,jsFutureKeys,jsDomErrNo,jsDomNodeConsts,jsHtmlEvents,jsFuncCall,jsUndefined,jsNan,jsPrototype,jsBuiltins,jsNoise,jsClassDefinition,jsArrowFunction,jsArrowFuncArgs,jsParensError,jsComment,jsArguments,jsThis,jsSuper,jsDo
```
amadeus: Since I've already mentioned above that I'd prefer not to have `jsConstant` as it exists now, this will would also need to be removed. As a general rule though, if you add things to a large contains, it's much easier if you add them to the end of the list because it makes handling merge conflicts much simpler.

jfelchner:
@amadeus all great feedback thanks!

Regarding the end of line semicolons, the look ahead was just for the comments. I'm fine with doing it a different way if you have another suggestion. I'm not sure how you would do it for each individual type of statement and be sure to grab it. But I'll give it a shot. :)

With regards to the constants, no, JavaScript does not _force_ this but have you ever been in a newer codebase (or any codebase) where this was not the convention?

I know with `const` people often create constants that start with a lowercase letter. That's not what I'm saying. What I'm saying is, have you ever seen a variable that starts with an uppercase letter that you didn't treat like a constant?

So I think that this is something that would be very useful to most developers. 

Potentially this could be added as an option like you're doing for concealing certain matches?

jfelchner:
I just realized I missed what you were saying on end of line semicolons. Matching on a particular type of semi wouldn't do what I wanted. :(

The issue is thus:

```javascript
var foo = 'hello';
var bar = 'goodbye';
```

Both semis in those lines should be matched. Because they're end of line semis, the have no use to the reader and are noise. They're there only for the compiler (minification, etc)

If I wrote it like this:

```javascript
var foo = 'hello'; var bar = 'goodbye;
```

Then only the second semi is noise in this case. The first on acts as a break between statement so that it's easier for the _human_ to parse. 

Obviously this is a contrived example.

amadeus:
Thing is though, the last semi in a line, to style it differently is still completely a stylistic thing, and even seemingly kind of arbitrary. Take an example like this:

example

I still think this is a case of you creating an `after/syntax/javascript.vim` file for JS that applies your stylistic additions and doesn't really belong in a generalized JS syntax file.

amadeus:
On the subject of consts, yes, while I agree it's pretty common, I still personally don't think it belongs since it's a stylistic thing, and starts to get into lexical parsing, which our syntax file isn't really equipped to do.

syntax/javascript.vim
```diff
@@ -42,12 +42,12 @@ syntax keyword jsModuleKeywords  contained import
 syntax keyword jsModuleKeywords  contained export skipwhite skipempty nextgroup=jsExportBlock,jsModuleDefault
 syntax keyword jsModuleOperators contained from
 syntax keyword jsModuleOperators contained as
-syntax region  jsModuleGroup     contained matchgroup=jsBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
+syntax region  jsModuleGroup     contained matchgroup=jsModuleBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
```
amadeus: I am fine with this, however it would be good to link it back to `jsBraces` at the bottom.

syntax/javascript.vim
```diff
@@ -42,12 +42,12 @@ syntax keyword jsModuleKeywords  contained import
...
 syntax match   jsModuleAsterisk  contained /*/
 syntax keyword jsModuleDefault   contained default skipwhite skipempty nextgroup=@jsExpression
 syntax region  jsImportContainer start=/\<import\> / end="\%(;\|$\)" contains=jsModuleKeywords,jsModuleOperators,jsComment,jsString,jsTemplateString,jsNoise,jsModuleGroup,jsModuleAsterisk
 syntax region  jsExportContainer start=/\<export\> / end="\%(;\|$\)" contains=jsModuleKeywords,jsModuleOperators,jsStorageClass,jsModuleDefault,@jsExpression
-syntax region  jsExportBlock     contained matchgroup=jsBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
+syntax region  jsExportBlock     contained matchgroup=jsExportBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
```
amadeus: Same comment here regarding the link to `jsBraces`

syntax/javascript.vim
```diff
@@ -212,6 +211,10 @@ syntax region  jsCommentClass       contained start=/\/\*/ end=/\*\// contains=j
 syntax region  jsCommentMisc        contained start=/\/\// end=/$/    contains=jsCommentTodo,@Spell skipwhite skipempty nextgroup=jsBlock extend keepend
 syntax region  jsCommentMisc        contained start=/\/\*/ end=/\*\// contains=jsCommentTodo,@Spell skipwhite skipempty nextgroup=jsBlock fold extend keepend
 
+" Decorators
+syntax match   jsDecorator                    "^\s*@" nextgroup=jsDecoratorFunction
+syntax match   jsDecoratorFunction  contained "[a-zA-Z_][a-zA-Z0-9_.]*" nextgroup=jsParenDecorator
```
amadeus: I know this was from how it was previously, but while you are fixing this, might as well convert these to regexes instead of leaving them as strings

syntax/javascript.vim
```diff
@@ -324,8 +324,8 @@ if version >= 508 || !exists("did_javascript_syn_inits")
   HiLink jsModuleKeywords       Include
   HiLink jsModuleOperators      Include
   HiLink jsModuleDefault        Include
-  HiLink jsDecorator            Special
-  HiLink jsDecoratorFunction    Special
+  HiLink jsDecorator            Tag
+  HiLink jsDecoratorFunction    Tag
```
amadeus: Not sure `Tag` is the right one for this? The vim help states the following for `Tag`:

```
	 Tag		you can use CTRL-] on this
```

syntax/javascript.vim
```diff
@@ -337,7 +337,7 @@ if version >= 508 || !exists("did_javascript_syn_inits")
   HiLink jsSwitchColon          Noise
   HiLink jsClassMethodType      Type
   HiLink jsObjectMethodType     Type
-  HiLink jsClassDefinition      jsFuncName
+  HiLink jsClassDefinition      Constant
```
amadeus: I also disagree with this - Technically a class is a function, and therefore should be linked to a function. We provide the `jsClassDefinition` group so you can change it to another color if you'd like.

syntax/javascript.vim
```diff
@@ -290,7 +290,7 @@ if version >= 508 || !exists("did_javascript_syn_inits")
   HiLink jsOperator             Operator
   HiLink jsOf                   Operator
   HiLink jsStorageClass         StorageClass
-  HiLink jsClassKeywords        Structure
+  HiLink jsClassKeywords        Keyword
```
amadeus: 👍

amadeus:
Anyways, overall I like some of these changes. Always nice to get finer granularity with various braces and parens. However I definitely disagree with the stylistic opinionated changes, and therefore I think those should be removed before merging.

You could always remove them to get the good stuff merged in, and we could bike shed in another issue afterwards.

syntax/javascript.vim
jfelchner: It does yes?

syntax/javascript.vim
jfelchner: 👍🏻

syntax/javascript.vim
jfelchner: Where did you find that documentation!? :D

amadeus: Do `:help syntax`, then just do a normal `/` search for `Tag`

syntax/javascript.vim
jfelchner: Did you read my comment in the commit for that?  I mentioned that, yes, technically it's a function, but that users who use this style are trying to gloss over that fact.  Otherwise they'd build the thing in a prototypical way.

syntax/javascript.vim
amadeus: Or be smart, and not dumb like me, and do `:help group-name`

syntax/javascript.vim
amadeus: Yeah I did read it. Think is, there's intention and use, and then there's syntax accuracy. The methodology we try to take here (and don't always do it perfectly, mind you) is to match it accurately, then leave it up to each particular dev to customize as appropriate. Therefore, we match it with function because that's what it is, but allow the flexibility for it to be changed by a dev later.

amadeus: Basically the idea being here - we aren't trying to be opinionated about things, just provide syntactically accurate matching (or at least as close as possible, given the limitations of trying to use a regexes to parse a language design for a lexical parser)

syntax/javascript.vim
jfelchner: How did you even get this diff above?

syntax/javascript.vim
amadeus: `jsModuleBraces` is not linked to anything right now.

syntax/javascript.vim
jfelchner: Oh.... you're looking at the merged diff and not each individual commit.

syntax/javascript.vim
jfelchner: Ok.  I can definitely put this in an `after` for sure.  My take however would be to match everyone's 80% case and let them change it to something other than if they wanted.

Most people who use classes are going to want the name of the class highlighted as a constant (I'd assume).  If they want it highlighted as a function, then _they_ could go into an `after` and modify it to be so.

Basically it just boils down to how many people do you want to have to modify the syntax highlighting to make it like what they want.

But I'll remove it to get the rest of this merged and add it to my `after`. 👍🏻

syntax/javascript.vim
jfelchner: Missed that one!  Fixing.

syntax/javascript.vim
jfelchner: @amadeus what do you think about a global variable to add in syntax highlighting for constants?  Like you do with `this`, `return`, etc.

amadeus:
Btw, something like

```
syntax match   jsConstant           "\<[A-Z][A-Za-z]\+\>" containedin=jsDestructuringBlock
syntax match   jsSemiEndOfLine      /\;\{1}\(\(\s*\/[\/\*].*\)\?$\)\@=/

syntax cluster jsExpression add=jsConstant,jsSemiEndOfLine

hi! def link jsConstant        Constant
hi! def link jsClassDefinition Constant
```

In `after/syntax/javascript.vim` should accomplish your stylistic changes.

syntax/javascript.vim
amadeus: Not sure I follow?

syntax/javascript.vim
amadeus: It's actually difficult to discuss what people want. I usually judge it based on issues and PRs that show up. Truth is, almost no one ever files issues about linking groups. Occasionally we get people who file issues that they want to change the color of something, at which point we just post an example of how they can add a special color that particular group, and they are on their way.

In other words, people generally don't care what things are linked too, they just want their colors to look good.

With that in mind, my vote will always go for Vim syntax accuracy, since we always have a match group name that can then be targeted or isolated.

In regards to class stuff, we've never actually had people file an issue about this, usually it's with regards to things like parenthesis and braces and stuff.

syntax/javascript.vim
```diff
@@ -41,12 +42,12 @@ syntax keyword jsModuleKeywords  contained import
...
+syntax region  jsModuleGroup     contained matchgroup=jsModuleBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
 syntax match   jsModuleAsterisk  contained /*/
 syntax keyword jsModuleDefault   contained default skipwhite skipempty nextgroup=@jsExpression
-syntax region  jsImportContainer start=/\<import\> / end="\%(;\|$\)" contains=jsModuleKeywords,jsModuleOperators,jsComment,jsString,jsTemplateString,jsNoise,jsModuleGroup,jsModuleAsterisk
+syntax region  jsImportContainer start=/\<import\> / end="\%(;\|$\)" contains=jsModuleKeywords,jsModuleOperators,jsComment,jsConstant,jsString,jsTemplateString,jsNoise,jsModuleGroup,jsModuleAsterisk
 syntax region  jsExportContainer start=/\<export\> / end="\%(;\|$\)" contains=jsModuleKeywords,jsModuleOperators,jsStorageClass,jsModuleDefault,@jsExpression
-syntax region  jsExportBlock     contained matchgroup=jsBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
+syntax region  jsExportBlock     contained matchgroup=jsExportBraces start=/{/ end=/}/ contains=jsModuleOperators,jsNoise,jsComment
```
amadeus: Also don't forget to link this to `jsBraces` as well.

syntax/javascript.vim
jfelchner: 👍🏻

syntax/javascript.vim
jfelchner: Something like this:

```vim
if exists('g:javascript_highlight_upper_camel_as_constants')
  syntax match   jsConstant           "\<[A-Z][A-Za-z]\+\>"

  syntax region  jsImportContainer    add=jsConstant
  syntax region  jsDestructuringBlock add=jsConstant
  syntax cluster jsExpression         add=jsConstant
  syntax region  jsDestructuringBlock add=jsConstant

  HiLink jsClassDefinition      Constant
  HiLink jsConstant             Constant
else
  HiLink jsClassDefinition      Function
endif
```

syntax/javascript.vim
jfelchner: Done

syntax/javascript.vim
amadeus: Hmmm, not sure yet. Let's move that to another issue or PR for further discussion. My initial thought is that it adds clutter where it doesn't need to be, but other maintainers here may have counter opinions and it would be good to isolate that discussion.

syntax/javascript.vim
amadeus: I don't see the changes in this PR yet? Mebe not pushed yet?

jfelchner:
@amadeus I think I got everything.  I'll open another issue with the constant discussion.

syntax/javascript.vim
```diff
@@ -74,12 +74,14 @@ else
 endif
 syntax cluster jsRegexpSpecial    contains=jsSpecial,jsRegexpBoundary,jsRegexpBackRef,jsRegexpQuantifier,jsRegexpOr,jsRegexpMod
 
+" Objects
 syntax match   jsObjectKey         contained /\<[0-9a-zA-Z_$]*\>\(\s*:\)\@=/ contains=jsFunctionKey skipwhite skipempty nextgroup=jsObjectValue
+syntax match   jsObjectColon       contained /:/ skipwhite skipempty
```
amadeus: Quick thing - we should link this back to `jsNoise` for backwards compat.

jfelchner: Done

syntax/javascript.vim
amadeus: We should figure this one out as well.

amadeus: I think `jsDecoratorFunction` should probable be linked to `Function` or `Identifier`, then `jsDecorator` should perhaps be linked to `Operator`, `Label`, `Type`, or `Special`

amadeus:
Outside of the `Tag` issues I mentioned before, I think this should be good to go. I will test by using it today for a bit, and if everything appears ok, then I'll merge this evening. The only reason I am waiting to test for a couple hours is we have a pretty big merge going out to master soon, and I don't know the timing of when it will get merged, and don't want this to inadvertently hold it back if we find a bug or issue.

jfelchner:
@amadeus you seemed ambivalent on the `Tag` thing so I left it in.  I think it still works.  `Tag` seems like an unused one.  Most of the time tagged items are also another item (class name, method name, etc).

All of those groups seem very specific to C IMO.  Not all of them apply to other languages.  Nor in the same way.

jfelchner:
Up to you.  I can add it to my local config.

amadeus:
I definitely thing `Structure` that we had from before was wrong. So it would be good to fix, it's a quick fix though, let me think about it for a bit.

amadeus:
Doh! Of course yesterday I was heads deep in Objective-C and had no time to test, hopefully today will be ok, or if Dev gets merged into Master, then we can just go ahead and test in dev first, but it's up to @bounceme for when he's ready.

bounceme:
I'm good with this going into the dev -> master merge when you guys are ready. no rush

amadeus:
This is merged, once again, thanks @jfelchner for the PR!

I went ahead and just added the hilinks into develop manually, to get this in.

PR #636, opened 2016-09-16
multi-braceless statements

bounceme:
very unrefined so far

This changes:
```
if (1)
  for (;;)
  i +=
  l
```
to:
```
if (1)
  for (;;)
    i +=
      l
```


bounceme:
there doesn't seem to be any bugs with this, I'll merge tomorrow. wish I could hone that function down though

PR #642, opened 2016-09-24
normalize header in vim and current, date

bounceme:
Any issues with this? It doesn't rlly mean anything, just something to satisfy vim/dev


amadeus:
No problem at all, Vim can be funky about these things. And you are essentially the only one working on it xD

PR #645, opened 2016-09-24
Fix jsFlowTypeCustom regex to allow dots

remiprev:
It’s possible to use dots in Flow types, but the syntax did not allow it, and it broke how the rest of the content was highlighted.

I’m not really sure why `\k` was used here (I don’t have a lot of experience with Vim’s `syntax match`), but I’ll gladly update the regex with something more appropriate, as long as it accepts dots 😄


amadeus:
Looks good, thanks!

remiprev:
Woah, that was fast. Thanks! 😀

PR #654, opened 2016-10-03
Add support for "for await" (asynchronous iteration)

BenoitZugmeyer:
That's my first edition of a vim syntax file, please review carefully.

[Asynchronous iteration](https://github.com/tc39/proposal-async-iteration) has recently been promoted to stage3, so its syntax is now quite stable and should land in ES as is.


bounceme:
can it be braceless?
```
for await (;;)
  something()
```

bounceme:
it must be, if for-of can. never mind

bounceme:
have any thoughts @amadeus ?

PR #659, opened 2016-10-05
Fix JSDoc highlighting for optional parameters

### LINKED ISSUE #574: JSDoc highlighting for optional parameters
I've noticed that `vim-javascript` does an excellent job with most JSDoc highlighting, but gets confused on [optional parameters](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values).

Here's a screenshot showing how both "normal" and "optional" parameters are handled.

screen shot 2016-07-26 at 1 29 54 pm

It would be great if it showed up the same color, and even better if the each part was highlighted separately to make the default value stand out, if provided.

tomxtobin:
Fixes #574.


tomxtobin:
Aw, crap. Botched that.

tomxtobin:
Post-mortem: the `hub` CLI tool automatically made the PR against `master` while I wrote it against `develop`, and I couldn't figure out how to get that unstuck after I force-updated the branch to resubmit it against `master`. Resubmitted it as #660.

PR #660, opened 2016-10-05
Fix JSDoc highlighting for optional parameters

### LINKED ISSUE #574: JSDoc highlighting for optional parameters

tomxtobin:
Fixes #574.

### DIFF
--- a/extras/jsdoc.vim
+++ b/extras/jsdoc.vim
@@ -18,7 +18,7 @@ syntax region jsDocTypeRecord   contained start=/{/ end=/}/ contains=jsDocTypeRe
 syntax region jsDocTypeRecord   contained start=/\[/ end=/\]/ contains=jsDocTypeRecord extend
 syntax region jsDocTypeNoParam  contained start="{" end="}" oneline
 syntax match  jsDocTypeNoParam  contained "\%(#\|\"\|\w\|\.\|:\|\/\)\+"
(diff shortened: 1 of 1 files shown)


bounceme:
`\|\[.\{-}]` to this `\|\[.\{-}\]`

tomxtobin:
> Does the closing bracket need to be escaped also?

Not with the default regex `magic`, no. (I find the magic rules so confusing that I always prefix my Vim regexes with `\v`, so they're fairly close to PCREs or POSIX EREs.)

bounceme:
oh, in the help it explains the behavior that some internal functions have:
```

								*E769*
	When the ']' is not there Vim will not give an error message but
	assume no collection is used.  Useful to search for '['.  However, you
	do get E769 for internal searching.  And be aware that in a
	`:substitute` command the whole command becomes the pattern.  E.g.
	":s/[/x/" searches for "[/x" and replaces it with nothing.  It does
	not search for "[" and replaces it with "x"!
```

bounceme:
Oh, that was into master haha. try not to do that

amadeus:
Oh whoops, it was from mobile (so I couldn't see the target), and since it wasn't brought up earlier, I just assumed it was correct.

bounceme:
I missed it too. the dev to master push should cover it up

PR #667, opened 2016-10-10
massive expression to if chain

bounceme:
more readable


bounceme:
Very nice improvement to performance also!

PR #691, opened 2016-10-29
Add fetch keyword to list of globals

### LINKED ISSUE #688: Highlight calls to "fetch" API
https://fetch.spec.whatwg.org/

Fetch is the promise-based XMLHttpRequest replacement. It's globally available such as other currently highlighted objects like `console` and `window`.

Example code taken from the MDN:
```javascript
fetch('flowers.jpg').then(function(response) {
  if(response.ok) {
    response.blob().then(function(myBlob) {
      var objectURL = URL.createObjectURL(myBlob);
      myImage.src = objectURL;
    });
  } else {
    console.log('Network response was not ok.');
  }
})
.catch(function(error) {
  console.log('There has been a problem with your fetch operation: ' + error.message);
});
```

mixedCase:
Closes #688.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -112,7 +112,7 @@ syntax keyword jsAsyncKeyword           async await
 syntax match   jsSwitchColon  contained /:/          skipwhite skipempty nextgroup=jsSwitchBlock
 
 " Keywords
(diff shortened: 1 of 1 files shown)


mixedCase:
@amadeus Pinging.

PR #696, opened 2016-11-07
Develop -> Master

bounceme:
the indent script is stable and I'm probably done working on the indent script at this point, except for bug reports.

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -50,7 +50,7 @@ function s:skip_func(lnum)
     return !s:free
   endif
   let s:looksyn = line('.')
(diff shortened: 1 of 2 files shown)


bounceme:
I may actually make a few changes to make do-whiles not break when there is a while in the do statement. that is the only bug I can think of

bounceme:
https://github.com/pangloss/vim-javascript/pull/515#issuecomment-231525547

https://github.com/mishoo/UglifyJS2/issues/893

I'm actually going to remove this support. All `while (/*do things*/) \n` are going to be indenpendent from do statements. This will simplify the code, and also spare me from attempting to accurately parse nested braceless `whiles` in `do's`

bounceme:
merging tomorrow. speak now or forever hold your peace

PR #701, opened 2016-11-08
master->develop

bounceme:
backport some fixes


amadeus:
I'd say we just delete develop and create anew from master, cleaner history that way?

bounceme:
Alright with me! My commits will be slowing a lot

PR #706, opened 2016-11-10
Implement jsFuncAssignment

### LINKED ISSUE #694: Highlighting identifiers as jsFuncName when used with arrow functions
To make identifying functions easier, it would be really nice if we could highlight identifiers that hold arrow functions as functions themselves.

For example,

image

`doSomething` is identified as `jsFuncName` whereas
`doSomethingElse` is identified as `jsVariableDef`.

The Babel ES6 grammar in Atom handles this, but I'm not sure if it has more capabilities than what we're able to do in Vim.

apazzolini:
This resolves #694, albeit in a potentially suboptimal / incorrect way (I am new to Vim syntax programming).

If you decide to merge this pull request, we may also want to remove the `HiLink jsFuncAssignment       Function` line as a default. 

I'm happy to accommodate any requirements for merging or alternatively keeping this in a private fork if this isn't functionality you think belongs in the main branch.

Thanks!

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -29,7 +29,7 @@ syntax match   jsFuncCall       /\k\+\%(\s*(\)\@=/
 syntax match   jsParensError    /[)}\]]/
 
 " Program Keywords
-syntax keyword jsStorageClass   const var let skipwhite skipempty nextgroup=jsDestructuringBlock,jsDestructuringArray,jsVariableDef
+syntax keyword jsStorageClass   const var let skipwhite skipempty nextgroup=jsDestructuringBlock,jsDestructuringArray,jsVariableDef,jsFuncAssignment
 syntax match   jsVariableDef    contained /\k\+/ nextgroup=jsFlowDefinition
 syntax keyword jsOperator       delete instanceof typeof void new in of
 syntax match   jsOperator       /[\!\|\&\+\-\<\>\=\%\/\*\~\^]\{1}/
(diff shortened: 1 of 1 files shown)


bounceme:
@amadeus decides in the end for syntax stuff.This works for me in terms of implementation, thanks

amadeus:
Soo, I am not actually a fan of doing this for a couple reasons.

Technically the thing we are identifying against, is not the variable/pointer to the function, but the actual 'name' of the function. Example:

example

Arrow functions are anonymous by nature. Function name and pointer can be used to execute the function when you have each defined.

This starts to get into a slippery slope with what our syntax file should support. We could highlight it on assignment, but never after assignment since we can never know what a variable has evaluated too. I am not a fan of that inconsistency.

I am going to close this for now, however the discussion absolutely can still continue.

bounceme:
apparently the [name property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) in es6 works on anonymous functions, though I am with @amadeus since the further references wouldn't work
```
let i = () => 1+1
let a = i
console.log(a.name) // logs i
```

afgomez:
@amadeus I kindly ask you to reconsider merging this.

The change this PR proposes mimics how most of the people uses JS arrow functions.

Consider this code:

```js
function foo1() { /* ... */ }
const foo2 = () => { /* ... */ }
foo2();

// You _can_ do this, but I personally don't think it needs highlighting.
const bar = foo2;
bar();

class SomeComponent extends ReactComponent {
  someMethod() { /* ... */ }
  someBoundEventHandler = () => { /* ... */ }
  render() {
    return <a onClick={someBoundEventHandler}></a>
  }
}
```

For `foo1` and `foo2` both identifiers have the same purpose. Having the same highlight signals this.

`const bar = foo2` is just a variable assignment. The developer knows `foo2` is a function, but `bar` is not where the function is _defined_. It's just a simple variable assignment.

This becomes more obvious in the class methods (I'm not sure this PR tackles that). The purpose of `someBoundEventHandler` is to be a function, but it's not highlighted as such.

This is how VSCode handles this code (theme https://github.com/afgomez/better-cobalt.vscode)

screen shot 2018-02-23 at 17 52 14


This is neovim with pangloss/vim-javascript (theme https://github.com/afgomez/better-cobalt.vim)

screen shot 2018-02-23 at 17 51 19

The VSCode version is more consistent with the purpose of each identifier.

---

PS: this is an awesome plugin :)

apazzolini:
For what it's worth, I've been using a fork of this repository with my changes for a long time now, and I haven't run into any issues. I last merged upstream on Jan 31, and it looks like there are 6 unmerged commits since then.

If you're considering pulling this into the main repo, I'd be happy to provide an up to date version.

@afgomez My fork doesn't handle a few edge cases, including what you described. It would definitely be nice if it did - I may attempt to add them, but my vim syntax programming skills are lacking.

apazzolini:
@afgomez I've updated my fork for object and class property arrow functions. 

I'm not sure if there are wide implications of my changes that I missed, but everything looks good on my codebases. If you try this out and see something broken, let me know and I can try to fix it. Here are the various places that get highlighting with my changes:



To the maintainers: I'm not necessarily advocating merging these changes into mainline - I totally get the rationale for not wanting to. Just leaving these comments here since it's the most likely place someone looking for this functionality would land.

afgomez:
@apazzolini wow, thanks a lot! I'll give it a try :)

PR #709, opened 2016-11-13
trim from start of comment sequence

bounceme:
this fixes the edge case:
```
code = 'Neat' /// comment, through trimmed to a '/'
```


bounceme:
this only works with vim's nfa reg engine, i'll file a bug report

PR #713, opened 2016-11-16
faster prevline

bounceme:
is just as fast as before and also far quicker with a file like: https://raw.githubusercontent.com/sanctuary-js/sanctuary-def/master/index.js


amadeus:
Any benchmarks that might show these improvements? Assuming they aren't a hassle to generate

bounceme:
with sanctuary-def:
before: 11.6 seconds
after: 4.6

that's the only file where the difference is so drastic though, most files are slightly faster or the same

bounceme:
I will merge this later, is anyone against this?

PR #715, opened 2016-11-18
Performance, searchpair optimization

bounceme:
This improves performance, especially when using the default syntax where the difference is massive

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -239,8 +239,8 @@ function GetJavascriptIndent()
   let num = b:js_cache[1]
   let stmt = exists('stmt') ? 1 : 0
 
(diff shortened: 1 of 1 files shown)


bounceme:
with the default syntax file i just indented all of phaser.js (104000 lines with massive functions) in 2.5 minutes

https://raw.githubusercontent.com/photonstorm/phaser/dev/build/phaser.js

bounceme:
this is flawed, as it doesn't account for operator first style. the isBlock function is used to determine whether the line is part of an expression: if it is, then we have to searchpair with `[{(` because the context could be a block, argument list, object, array, etc; unfortunately when on a empty line this breaks down as we can't determine whether the user just needs to continue a expression, by typing an operator

Issue #723, opened 2016-11-21
Issue with with flow type object


gf3:
In flow, object keys can be terminated with a semi-colon (vs a comma in JS) and that seems to be tripping up syntax highlighting:

<img width="465" alt="screen shot 2016-11-21 at 10 57 19 am" src="https://cloud.githubusercontent.com/assets/18397/20489643/5b5aefca-afd9-11e6-98b0-17fe99cacde4.png">

This is with version a87c94438b038a060fa2920055a50a0c1cf08186

amadeus:
I've recently created a branch with some flow fixes, if you could do me a solid and test it: https://github.com/pangloss/vim-javascript/pull/991

PR #724, opened 2016-11-21
Add `jsFlowObjectValue` syntax region

gf3:
Partial fix for #723, still having issues grabbing that last semi-colon as either `jsNoise` or `jsFlowNoise` but the type definition looks a lot better:

screen shot 2016-11-21 at 11 14 34 am


bounceme:
Isn't that just for the property annotations?
```
// @flow
class Foo {}
var obj: {a: boolean; b: string; c: Foo} = {a: true, b: "Hi", c: new Foo()}

```

bounceme:
unless all of the flow syn-names are specifically about the annotations

amadeus:
This PR will overwrite `jsObjectValue` object definitions as `jsFlowObjectValue` which can cause styling problems.

Let me think about this one for a bit and see if I can come up with something.

bounceme:
why not: match a object, then lookback two words for `type` ?

Issue #683, opened 2016-10-23
Indentation issue


sassanh:
Is it an issue or am I missing something:
```
if (false ||
    true) {
        do_something();
    }
```

Last 2 lines seems to have extra indentation.

bounceme:
https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244585305

 you've happened to ask that like 4 times now

sassanh:
I'm absolutely sure it was fixed.
Btw you said here https://github.com/pangloss/vim-javascript/pull/621#issuecomment-244530847:

thats unrelated, lines that start with }]) are indented to the line which contains the other pair. wontchange

Then I guess we should change the logic. The result is just no right, C++ indenter in vim doesn't indent like this too:
C++:
```
int main() {
    if (true and
            false) {
        cout << 123;
    }
}
```
If `}` indents to the other pair then I guess we should just change it. Tell me you don't have time but you agree with me and I'll manage to do this or maybe we decide to just ignore it. But "it's correct the way it is just because it's our current implementation" doesn't make sense.

sassanh:
@bounceme 
Btw, I'm sorry if my issues annoy you. I'm absolutely thankful for your job here. If it's the case that you find these issues irrational or annoying just tell me and I'll try to find another way for my problems.

bounceme:
It's correct. I would keep the behaviour as it is even if someone submitted a p.r. 
```
int main() {
  if (true and
    false) {
      cout << 123;
    }
}
```
this is our behaviour and I find it easier to understand. Also the possibility of this being added without serious performance loss is slim

bounceme:
https://github.com/pangloss/vim-javascript/pull/684

bounceme:
I still don't like it compared to the current behaviour though. It is a matter of preference and code complexity

bounceme:
> I'm absolutely sure it was fixed.

Would I not notice making this 'fix' myself?

sassanh:
Thanks for that commit, it fixes the brace but not the content.
OK, what about a setting variable for it? So that user can decide to have it the way he prefers based on preference and his machine's resources. Even though it's hard to understand for me why you prefer it the way it is, I guess you agree that the one I'm looking for is a valid at-least-good indentation that maybe less or more around %50 of users prefer it that way. Maybe it deserves a setting variable.

> Would I not notice making this 'fix' myself?

[...]

PR #725, opened 2016-11-24
paren start line as baseline in blocks

bounceme:
https://github.com/pangloss/vim-javascript/issues/683#issuecomment-255620211

@sassanh what do you think about this ? list-like things (arrays, ob literals, argument lists) will all now be indented extra when using operators besides the comma, and grouping parens are treated blocklike


sassanh:
Sounds great. I tried it in few files and it seems to be an improvement from my point of view.

bounceme:
hey, I'm going to simplify this. I figured out what we want, i think, and what jsbeautify does

sassanh:
This new version is working alright for me. Good job.

PR #729, opened 2016-11-25
operator cleaning

bounceme:


bounceme:
need to switch the changes to the other patterns hah

PR #730, opened 2016-11-25
Add '0.' to 'indentkeys'

adriaanzon:
This will automatically indent when chaining methods on a new line.

__Example of when this is useful:__
You have this piece of code, where `|` represents the cursor:

```javascript
this.$http.get('/foo')
|
```

You type `.then(`

```javascript
this.$http.get('/foo')
    .then(|
```

The line got reindented when pressing `.`

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -13,7 +13,7 @@ let b:did_indent = 1
 " Now, set up our indentation expression and keys that trigger it.
 setlocal indentexpr=GetJavascriptIndent()
 setlocal autoindent nolisp nosmartindent
-setlocal indentkeys=0{,0},0),0],:,!^F,o,O,e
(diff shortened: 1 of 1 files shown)


bounceme:
it would be useful with every operator.it can make editing slower though.

bounceme:
maybe we should just change our setting to use `+= or ^=` so that the user's options aren't overridden

adriaanzon:
Currently I just use this:

```vim
autocmd FileType javascript setlocal indentkeys+=0.
```

which is also fine.

bounceme:
I actually think we are using the default global indentkeys setting. maybe we can just remove it altogether

bounceme:
i hope you don't mind but i think this is better as a user configuration ( maybe we can add a note here: https://github.com/pangloss/vim-javascript#indentation-specific ) . I really hate when plugins change things like indentkeys and formatoptions

PR #731, opened 2016-11-25
Completely refactored how import/export work

amadeus:
I've never liked how I originally built import/export. This refactors everything keeping the logic simpler and also share matches between them.  Using regions I think was a mistake on my part, I've opted instead to use nextgroups which works much better.

Basically this was inspired because of yesterdays bug, and seeing the code I had written originally 😄

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -37,16 +37,14 @@ syntax keyword jsBooleanTrue    true
 syntax keyword jsBooleanFalse   false
 
 " Modules
-syntax keyword jsModuleKeywords  contained import
-syntax keyword jsModuleKeywords  contained export skipwhite skipempty nextgroup=jsExportBlock,jsModuleDefault
-syntax keyword jsModuleOperators contained from
-syntax keyword jsModuleOperators contained as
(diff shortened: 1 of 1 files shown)


amadeus:
This is the test file I am using for this PR, everything here works, I think this has all the various possible cases:

```js
import Module from 'module';
export const exported = {};
export { name1, name2, nameN };
export { variable1 as name1, variable2 as name3, nameN };
export let name1000, name2000, nameN000;
export let name10000 = 2, name200000 = 3, nameN000000;

export default expression;
export default function () {}
export default function name1() {};
export { name1 as default};

export * from 'module';
export module from 'module';
export { name1, name2, nameN } from 'module';
export { import1 as name1, import2 as names2, nameN } from 'module';

import defaultMember from "module-name";
import * as name from "module-name";
import { member } from "module-name";
import { member as alias } from "module-name";
import { member1 , member2 } from "module-name";
import { member1 , member2 as /* another comment */ member2 } from "module-name";
import {
	member1,
	// A comment
	member2 as alias2,
} from "module-name";
import defaultMember, { member as example } from "module-name";
import defaultMember, * as name from "module-name";
import "module-name";
```

PR #735, opened 2016-11-27
Added conceal for no-argument arrow functions - either `()=>` or `_=>`

hash-bang:
Setting `let g:javascript_conceal_noarg_arrow_function` to a character value conceals the characters before an arrow function. The syntax matcher also checks for whitespace after the `()` or `_` characters.

screenshot_2016-11-27_11-25-41

In the above screenshot the top half of the window shows the raw file and the bottom half shows the VI output when the following style is present in `~/.vimrc`:

```vimrc
let g:javascript_conceal_arrow_function = "🡆"
let g:javascript_conceal_noarg_arrow_function = "🞅"
set conceallevel=1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ global vim option, we do not set it ourselves.
 ## Concealing Characters
 
 You can customize concealing characters by defining one or more of the following
-variables:
+variables (these are all disabled by default):
 
     let g:javascript_conceal_function             = "ƒ"
     let g:javascript_conceal_null                 = "ø"
@@ -98,6 +98,16 @@ variables:
     let g:javascript_conceal_arrow_function       = "⇒"
     let g:javascript_conceal_noarg_arrow_function = "🞅"
     let g:javascript_conceal_underscore_arrow_function = "🞅"
+
+
+You can enable concealing within VIM with:
+
+    set conceallevel=1
+
+OR if you wish to toggle concealing you may wish to bind a command such as the following which will map LEADER+L (leader is usually the comma key) to toggling conceal mode:
+
+    map <leader>l :exec &conceallevel ? "set conceallevel=0" : "set conceallevel=1"<CR>
+
 
 ## Indentation Specific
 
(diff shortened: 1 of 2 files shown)


bounceme:
`_` being a 'black hole' argument?

bounceme:
I think the currying people would be a little unhappy with that option set

hash-bang:
I suppose its a matter of style.

I could move it to another optional setting if you think this keeps everyone happy?

bounceme:
that'd be better. I'd be happy with it being merged

hash-bang:
There you go.

By the by, you may also want to tweak your README a little to add something along the following lines:

> To enable the conceal use `set conceallevel=1`
> You can set up a keymap to toggle conceals using `leader-key + l` with `map <leader>l :exec &conceallevel ? "set conceallevel=0" : "set conceallevel=1"<CR>`


... as its not obvious this is how you actually turn it on. I wasn't sure if I should submit a patch to the README for this or whether its just better to tell you.

bounceme:
so is the underscore argument a common thing at the moment? I haven't seen that before as being a pattern many use. About the readme: i don't use the concealing at all so have no idea really

hash-bang:
I think its just a taste thing really. My team and I much prefer `_=>` as it just 'looks better', although I realise this isn't exactly a reasoned argument. There is some [discussion here](http://stackoverflow.com/questions/33427800/empty-parentheses-vs-underscore-for-es6-no-args-arrow-functions) but I suspect this is one of those 'space vs. tabs' style holy wars.

Fair enough about not using conceals but to be honest but that's what attracted me to your plugin the first place. I guess everyone has their own preferences. Would you like me to amend your README with what I posted?

bounceme:
tabs and spaces don't change the meaning of your code, but I understand it may be common enough to use the underscore. If you want to add a small note about concealing I would gladly merge it,though should probably be a new pr

hash-bang:
There you go

bounceme:
thank you very much. @amadeus approves?

PR #764, opened 2016-12-21
cleaning, edge case with keyword

bounceme:


bounceme:
should probably remove these ghetto switch expressions: https://github.com/pangloss/vim-javascript/blob/9fbfadc04c1ad3c1299711d33e495b2812a9e76f/indent/javascript.vim#L175

PR #774, opened 2016-12-25
fix async and static highlighting

k33nice:
Hi
I think current highlighting for ```async``` slightly confusing.
image

It seems to me that it is easier to read, when it identical highlighting in all cases.
image

Also, ```static``` keyword it's not allow to use into object literals, AFAIK.


amadeus:
Thanks for the heads up! I decided to just fix this in a separate PR to re-use more of what we already had.

Issue #600, opened 2016-08-16
labelled blocks


bounceme:
Wouldn't  be a very high priority bug as i don't think any js dev actually uses these but this breaks highlighting (and subsequently indentation)

```
outer_block: {
  inner_block: {
    console.log('1');
    break outer_block; // breaks out of both inner_block and outer_block
    console.log(':-('); // skipped
  }
  console.log('2'); // skipped
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/break#Examples

amadeus:
This may be a nightmare to support :(

Or maybe not, because I created a bad ass new syntax matching system. I would prefer to hold off on implementing until we start getting people complaining (or if one of us is starting to use it in the workplace xD)

bounceme:
Sounds good to me, unless we see proof of these ever being used just leave this open

romgrk:
Proof
```js
  here: for(var idx in salesReps) {
    var territories = salesReps[idx].territories;

    if (!territories)
      continue here;

    territories = territories.split(',');

    for(var tx in territories) {
      if(territories[tx] == territory) {
        salesRep = salesReps[idx].user;
        break here;
      }
    }
  }
```

bounceme:
I see labelled loops occasionally, but those already work here right? I meant just a block statement

bounceme:
Oh, the indent logic was actually removed in one of my refactorings

romgrk:
Doesn't seem to have any syntax applied on my setup; though I admit I'll use them like once a year.

bounceme:
Close this for #687 ?

amadeus:
Labelled block support is currently being tested in #778 

Assuming no major issues are found in the next few hours, it will get merged into master.

amadeus:
This feature is now in `master`

PR #778, opened 2016-12-26
Feature: Block Statement & Label Support

### LINKED ISSUE #687: wrong highlight with ES6 block scope
In such case, highlights and indentations do not work:

``` js
{
  function a() {}
  a();
}
```

image

amadeus:
This is still very rough, and only a first pass. There are some pretty major changes here and so I want to sit on this a bit and test it out. Others can feel free to give this branch a try as well.

I currently have not yet added support for labels, but I want to see if this direction I took has legs to it and doesn't break a bunch of things.

Theoretically, when/if this gets merged, it should fix #687 and #600,

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -30,9 +30,9 @@ syntax match   jsParensError    /[)}\]]/
 
 " Program Keywords
 syntax keyword jsStorageClass   const var let skipwhite skipempty nextgroup=jsDestructuringBlock,jsDestructuringArray,jsVariableDef
-syntax match   jsVariableDef    contained /\k\+/ nextgroup=jsFlowDefinition
-syntax keyword jsOperator       delete instanceof typeof void new in of
-syntax match   jsOperator       /[\!\|\&\+\-\<\>\=\%\/\*\~\^]\{1}/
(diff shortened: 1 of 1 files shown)


bounceme:
great, with `let` I think blocks can be somewhat useful now. Labelled blocks are a edge case though

amadeus:
@bounceme I agree, however the label support should be pretty trivial to add, once this in place and not causing any regressions.

amadeus:
Aright, this branch should be ready, minus a bit more testing.

Block statements and labels should be supported.

I just added labels a bit ago, so will sit on it for at least another day to be sure, however the original implementation I had appears to be pretty sound and besides one small ternary issue I found, has been bug free.

bounceme:
edge case:
```
switch(1){
  case 1 ? 2: a:
    i = 'sdfsdf'
}
```
`a` highlights as `jsBlockLabel` instead of `jsSwitchCase`

amadeus:
@bounceme fixed! good find

amadeus:
Alright, I think the best way to get this tested, is going to be getting it into `master` and let the potential bug reports roll in. For the most part, I have been using this branch for about a week or so and not noticed anything. So hopefully it's just a couple small edgecases to solve, or hopefully it's perfect as is!

PR #788, opened 2016-12-30
ternary function

bounceme:
very rough at the moment, maybe can replace `s:jump_label()` as well


bounceme:
@amadeus what are the syntax names that are used for `: or ?` in flow? so they can get skipped when parsing ternary operators

bounceme:
as long as it won't shadow ternary ops i can just match `jsflow`

bounceme:
1b6e19d also gives perfect label recognition!

PR #791, opened 2017-01-04
Bugfix for jsArrowFuncArgs

amadeus:
If the parens in a jsArrowFunction definition contained no arguments,
they were not assigned the `jsArrowFuncArgs` highlight group, which is
technically a bug.


hash-bang:
@amadeus - just pulled in this patch and it seems to break the desired behaviour of concealing empty arrow function syntax (i.e. `()=>`)

The original reason for the line that was added is in #735 which was to show empty arrow functions. Your change seems to negate this into _never_ showing them.

Unless I'm missing something?

amadeus:
This change is to ensure consistency with what we match.

Before this PR, parenthesis in an arrow function that were empty were not matched to the same highlight group as parenthesis that contained variables, this is a bug.

I.E. if someone wants to make the parenthesis of an arrow function a particular color, it would only work with variables defined between the parenthesis.

Thus this PR and fix.

hash-bang:
Fair enough on the coloring but the fix appears to break the conceal functionality.

The culprit seems to be this suffix: `.(' contains=jsArrowFuncArgs')` of line [176 in your patch](https://github.com/pangloss/vim-javascript/commit/2c44678d3e7e7cfce922764a46188eb7b564d1b6#diff-af5b29be69c5b34c00e40f13b7e92435L176)

This is a shot of the rendering with it removed:

screenshot_2017-01-12_16-59-57

And here with the prefix:

screenshot_2017-01-12_16-59-19

I'm guessing that the 'contains' matcher is upsetting something?

jdsutherland:
Ping @amadeus.

As @hash-bang says, this makes `let g:javascript_conceal_noarg_arrow_function` conceal no work, so it's currently dead code.

PR #803, opened 2017-01-12
fix HiLink type and add some class links and funtion links

kaidiren:
with this pr: the vim auto-complete tools will wokd better on some words. 😀


amadeus:
Can you give some examples regarding what would get fixed with this PR? It looks like it mostly just re-organizes a few items.

PR #808, opened 2017-01-14
add compiler plugin

bounceme:
adds eslint as a compiler for javascript files. It is used like:
`:compiler eslint` then `:make`
Most people already use plugins like neomake or syntastic, and i don't think this should be anything but a sensible fallback, i.e. not really interested in p.r.s adding different linters/styleguides etc.


bounceme:
@amadeus thoughts?

bounceme:
since vim has no compiler file for js i might submit this to Bram also

bounceme:
since we even have ctags definitions (by the way, how useful are those?), i don't see any problem with this

PR #814, opened 2017-01-16
First pass at jsObjectProp

amadeus:
I had an interesting idea to disable various matches like `class` and `require` from appearing when using `.` notation on objects (e.g. in `obj.class`, the `class` prop should not match as a class definition).

The proposed fix is to create a special `jsObjectProp` match, with a super high priority, that is used as a `nextgroup` for `.`

It is not assigned any colors or match groups, it simply exists to disable other items from matching.

This is a very quick pass and I have not really tested it yet, but I think it could be a very viable, working solution

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -24,7 +24,8 @@ syntax sync fromstart
 " syntax case ignore
 syntax case match
 
-syntax match   jsNoise          /[:,\;\.]\{1}/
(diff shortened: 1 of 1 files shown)


marlun:
Started using it and first tests seems fine, will use it today and get back if there's any problems.

meznaric:
Same as @marlun, what was broken before seems to be fixed. I did not notice any regressions yet.

PR #823, opened 2017-01-24
shorten final return

bounceme:


qstrahl:
Can you elaborate on what this change is for?

bounceme:
basically removes one ugly condition, keeping the same results and reusing the "offset" pattern: https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L314

qstrahl:
Ah, gotcha. LGTM :+1:

PR #830, opened 2017-01-29
template inner code

bounceme:


bounceme:
#830

marlun:
Works great, thanks! :)

PR #831, opened 2017-01-29
decimal precision?

bounceme:


bounceme:
i guess if someone wants this kind of thing
```
// sw=8 cino=:.25s
switch (1){
  case {e:3}:
          return 1 +
                  2
  case 1 + 1 ? 2: 2:
          return 1 +
                  2
}
```

PR #834, opened 2017-01-30
Enabling spell checking on template strings

mojir:
I have for some time been a bit annoyed that miss spelled words in template strings have not been highlighted.

The fix I did seem to fix this. However I don't know if I have broken something else


bounceme:
sorry for the lateness

PR #851, opened 2017-02-16
ternary experiment

bounceme:
very experimental, goal is better performance

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -133,7 +133,7 @@ function s:difcol()
         endif
       elseif s:looking_at() =~ '[;{]'
         return ':label'
-      elseif s:looking_at() == ':'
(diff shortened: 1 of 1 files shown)


bounceme:
i've tested this locally, will merge later. much faster than the old method

amadeus:
Any numbers on the improvements?

bounceme:
kind of difficult to say, since the old version was split into different functions.

https://raw.githubusercontent.com/sanctuary-js/sanctuary-def/779fc342b04f5d74778ec628a0b9bb80b204c8e5/index.js

old: 4.12s

new: 3.9s

http://pastebin.com/raw/eQ4cDaKf

old: 0.5s

new: 0.45s

bounceme:
there actually is some optimization left too

PR #876, opened 2017-03-09
reimplement Trim()

bounceme:
uses older version, only, as accurate as the current and faster than what is current


bounceme:
actually the version in master is bugged

PR #878, opened 2017-03-10
expr_col switch?

bounceme:
maybe this is even uglier ? any opinions welcome !


bounceme:
i can't merge this with a good conscious, without someones opinion. haha

bounceme:
i happen to like this better than the if chain

PR #883, opened 2017-03-12
refactor Prevcodeline

bounceme:
just trying this out


bounceme:
preferring traditional methods at the moment

Issue #905, opened 2017-04-01
highlighting for script in html is broken


gera2ld:
Open an HTML file with content below:
``` html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title></title>
</head>
<body>
  <script>
    import something from 'something';
    console.log(something);
  </script>
</body>
</html>
```

When not using any plugin, highlighting works for the embedded script. But when adding `pangloss/vim-javascript` plugin, `import` cannot be recognized and indention after `import` is broken.

Without `vim-javascript`:
![](https://cloud.githubusercontent.com/assets/3139113/24574394/6043b808-16c4-11e7-8ca7-796bd4b7122e.png)

With `vim-javascript`:
![](https://cloud.githubusercontent.com/assets/3139113/24574430/071d824e-16c5-11e7-8929-43a871cd5d9e.png)

bounceme:
i suggest you split it into a different file, or file a bug report with the html script maintainer. We have no control over html files.

gera2ld:
It only happens when this plugin is enabled, and the syntax highlight is quite different before and after the plugin is enabled (Just like that in a JavaScript file).
I have just looked into the HTML plugin, it does nothing with JavaScript, but just fallbacks to `indent/javascript.vim`, which in my case is from this plugin.

I have found the solution by appending `jsImport` and `jsExport` to `jsAll`, like this:
``` viml
syntax cluster jsAll contains=@jsExpression,jsStorageClass,jsConditional,jsRepeat,jsReturn,jsStatement,jsException,jsTry,jsAsyncKeyword,jsNoise,jsBlockLabel,jsImport,jsExport
```

gera2ld:
@bounceme Is there a reason to exclude `jsImport` and `jsExport` from `jsAll`?

bounceme:
Don't know, @amadeus might (I don't have any idea how syntax files work) . If you make a pr things will likely move faster with this

PR #906, opened 2017-04-01
fix: add jsImport and jsExport to jsAll

gera2ld:
ref: #905
 
Only `@jsAll` items will recognized in nested JavaScript blocks (such as HTML or VUE files).
Currently `jsImport` and `jsExport` are not included in `jsAll`. By adding them back, it works.


amadeus:
Technically we shouldn't put `import` in `jsAll` because that means you could theoretically put an `import` statement inside a function, they have to be a the top level of a a file.

The real fix for this, would be to append `jsImport,jsExport` to the `htmlJavaScript` cluster, after `@jsAll`

gera2ld:
Updated. Thanks.

PR #911, opened 2017-04-04
auto load flow file with comment header

bounceme:


bounceme:
hope this is alright with @amadeus

PR #913, opened 2017-04-05
move flow detect to ft detect file

bounceme:


bounceme:
also cleans up the ftdetect file patterns slightly. @amadeus is it alright to source the flow syntax after the regular syntax file has been sourced completely?

syntax/javascript.vim
```diff
@@ -233,7 +233,7 @@ if exists("javascript_plugin_jsdoc")
   endif
 endif
 
-if exists("javascript_plugin_flow") || search('\v\C%^\_s*%(//\s*|/\*[ \t\n*]*)\@flow>','nw')
+if exists("javascript_plugin_flow")
```
amadeus: Sorry I totally missed this back when you made the PR.

So I think this is causing the Flow file to source twice. Also, the way I think about it - Flow is not a file-type feature, it's actually a feature within javascript, so imo it should all still be within the syntax file. It also makes it consistent with the other language features we have.

I can work on a PR to clean this up?

Issue #567, opened 2016-07-21
[flow] defining a function that returns another function


BlooJeans:
On the latest develop (c7dca1bd6afdf900fc99e4636b49edffcde26af4)

test0: Correct (but no flow definition)
test1: Defining a function that returns another function w/ arrow function
test2: Defining a function that returns another function w/ regular function

<img width="247" alt="screen shot 2016-07-20 at 16 37 11" src="https://cloud.githubusercontent.com/assets/1751182/17002248/7356766e-4e98-11e6-91c8-e220c59219f1.png">

```
function test0(){
  return (ret) => {
    if (ret) {
      console.log('ret'+ret);
    }
    return;
  };
}

function test1(): (ret: any) => void {
  return (ret) => {
    if (ret) {
      console.log('ret'+ret);
    }
    return;
  };
}

function test2(): (ret: any) => void {
  return function(ret) {
    if (ret) {
      console.log('ret'+ret);
    }
    return;
  };
}
```

guyfedwards:
Having a similar issue, only with arrow functions however.
arrow function
![image](https://cloud.githubusercontent.com/assets/1794071/17729084/4d6b9d94-645a-11e6-8d02-c2d091ce9d1c.png)

`function` keyword
![image](https://cloud.githubusercontent.com/assets/1794071/17729093/5941f69a-645a-11e6-8e27-466ba38139eb.png)

anilanar:
Currently, using return types completely breaks syntax highlighting in function blocks:

I've used [this](http://vim.wikia.com/wiki/Identify_the_syntax_highlighting_group_used_at_the_cursor) to display current highlighting on the return flow type (you can see it in the status line).

<img width="1416" alt="screen shot 2016-12-07 at 00 35 36" src="https://cloud.githubusercontent.com/assets/753736/20948589/1c092eca-bc15-11e6-9140-30bfa816886d.png">

bounceme:
<img width="447" alt="screen shot 2016-12-06 at 5 42 18 pm" src="https://cloud.githubusercontent.com/assets/9332107/20951578/73f1002a-bbdb-11e6-859f-bf803e89a5cc.png">

you read the readme?

bounceme:
https://github.com/pangloss/vim-javascript/issues/567#issuecomment-240343110

this an issue for flowtype

amadeus:
[...]

PR #914, opened 2017-04-05
Allow flowtype function type as function return val

sebcode:
Fixes the test cases in #567


amadeus:
I believe I've sorta fixed this in a slightly different way in #991 

Can you give that branch a try? I could potentially improve it a bit more to mimic this PR more closely, but I am trying to work in a bunch of fixes all at once.

sebcode:
Awesome. I'll try your branch.

PR #931, opened 2017-05-02
&cino non-regex

bounceme:


bounceme:
shit, maybe i should just viml this:

```
    for (p = buf->b_p_cino; *p; )
    {
	l = p++;
	if (*p == '-')
	    ++p;
	digits = p;	    /* remember where the digits start */
	n = getdigits(&p);
	divider = 0;
	if (*p == '.')	    /* ".5s" means a fraction */
	{
	    fraction = atol((char *)++p);
	    while (VIM_ISDIGIT(*p))
	    {
		++p;
		if (divider)
		    divider *= 10;
		else
		    divider = 10;
	    }
	}
	if (*p == 's')	    /* "2s" means two times 'shiftwidth' */
	{
	    if (p == digits)
		n = sw;	/* just "s" is one 'shiftwidth' */
	    else
	    {
		n *= sw;
		if (divider)
		    n += (sw * fraction + divider / 2) / divider;
	    }
	    ++p;
	}
	if (l[1] == '-')
	    n = -n;
```

amadeus:
What these changes all about?

bounceme:
for cinoption parsing, using a hopefully clearer parsing method which is like how vim does internally ( that c code is what vim uses ). i was trying forever using regexes+substitution+eval, which is really confusing and fragile, because the option can be set incorrectly, causing the result to be something ridiculous or even harmful ( imagine your code disappearing, what actually happened is it is 2147483647 spaces to the left, because somehow we divide by 0, etc ).

Issue #866, opened 2017-03-03
Comment indentation wrong, also messes up anything after it


tvenable-mie:
Hi! I'm having a problem with indentation in comments. I am fully up-to-date with master. Here's an example:

**BEFORE INDENT**

```
/*
 * test
 */

function foo() {
	return 1;
}
```

**AFTER INDENT** (gg=G)

```
/*
* test
	*/

	function foo() {
		return 1;
	}
```

So you can see, not only is the comment itself wrong, but everything after that gets indented too many spaces. Thanks for looking!

bounceme:
can't reproduce, only guess is that you have syntax highlighting turned off. otherwise it's your config

tvenable-mie:
With a brand new file in that same session, I was getting the issue... but after closing Vim and going back to that file, everything was working again. Something must've gotten messed up in my session. I'll keep that in mind for the future. Sorry for the noise!

amadeus:
So just to add to this, I have yet to find a way to reproduce, but occaisionally I do find that the indentation just seems off and refuses to correct itself. As if there is some cached variable that wasn't updated or something. This always occurs in long coding sessions. If I open a new Vim instance and try to recreate it, the problem goes away

bounceme:
I know the cause @amadeus , it is an optimization unfortunately. https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L325

The `top` variable, which sets the limit to paren searching to the prev line if it had non-string chars in column 0. This is only a problem until the https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L321 condition is met (consecutive lines indented, like normal editing)

PR #933, opened 2017-05-03
break search at col:1

bounceme:
https://github.com/guns/vim-sexp/blob/master/autoload/sexp.vim#L126

main difference is that we break whenever we are beyond a col:1 `[\[\](){}]`


bounceme:
this will fix https://github.com/pangloss/vim-javascript/issues/866#issuecomment-288335316

PR #941, opened 2017-05-09
Fix buffer variables inited only once

yuezk:
@bounceme I noticed you changed the variable `s:syng_strcom` and `s:syng_str` to buffer variable in [this commit](https://github.com/pangloss/vim-javascript/commit/896bb54d2e2179258cd5da4290692faa1109a06e#diff-564c1f6e79518959361f7ad8a86f5792L63), which cause the buffer variable only init once. Because when there was a `*GetJavascriptIndent` definition, the code will exit. and the buffer scope variable won't init.

This may cause some weird indent in javascript.

### DIFF
--- a/indent/javascript.vim
+++ b/indent/javascript.vim
@@ -25,6 +25,17 @@ setlocal indentkeys+=0],0)
 
 let b:undo_indent = 'setlocal indentexpr< smartindent< autoindent< indentkeys<'
 
+" Regex of syntax group names that are or delimit string or are comments.
+let b:syng_strcom = get(b:,'syng_strcom','string\|comment\|regex\|special\|doc\|template\%(braces\)\@!')
+let b:syng_str = get(b:,'syng_str','string\|template\|special')
+" template strings may want to be excluded when editing graphql:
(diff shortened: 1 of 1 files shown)


bounceme:
it is used dynamically, and it can be changed whenever because it is b:scoped

bounceme:
so, i don't really see the point in the variables being redefined

yuezk:
Is [this](https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L234), [this](https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L256) and [this](https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L365) dynamically, too? I used `echom` to print the value, they are all empty when I switch between buffers.

bounceme:
yes. you have to edit a js file for them to be inited

bounceme:
oh shit, now i get it, can you move the s: vars below the `finish`

bounceme:
or we can make them g: scoped

bounceme:
merging

PR #942, opened 2017-05-09
fix #940: ternary expression returning arrow

### LINKED ISSUE #940: Syntax error when arguments to a function contain a ternary operator
And the ternary operator either calls a function or returns an arrow function.

In the following snippet, func2's closing parenthesis and func1's closing curly brace are wrongly marked as syntax errors.

```javascript
const func1 = function() {
  func2(func3 ? func3() : c => c)
}

// vs. (no syntax errors):
const func1 = function() {
  func2(func3 ? func3() : (c) => c)
}

// The following snippet is also free of syntax errors:
const func1 = function() {
  func2(func3 ? func3 : c => c)
}
```

bounceme:
#940 @amadeus how does this look? i can imagine it might have been intentional like that for flow or something else


amadeus:
Yeah, the flow stuff is a good point. I was thinking about regressions and all that, ugh.

amadeus:
However, typically when flow breaks something, hopefully it just breaks it in a way that still kinda works, so this may be ok. I'll sit a bit more on this patch and see

PR #943, opened 2017-05-10
Reuse pos restorer, misc cleaning

bounceme:


bounceme:
should also improve the naming of all the functions

bounceme:
whatever, i'm the only person who sees this file really. I'm used to the inconsistent names hah

bounceme:
tested on vim 7.2 - nvim 0.2

Issue #945, opened 2017-05-10
Wrong synID


sassanh:
Look at this line:
```
_.map(_.filter(this.state.data.chapters, foo => true), bar => {});
```
when cursor is on foo syntax is: ['jsParen', 'jsArrowFuncArgs', 'jsFuncArgs']
when cursor is on bar syntax is: ['jsParen', 'jsArrowFuncArgs']
I guess the second is correct, and first is getting wrong highlight due to wrong syntax id.

amadeus:
Oh yeah, something super messed up there, I'll take a look, thanks for the reproducible code snippet!

amadeus:
Alright, I have a branch in place with a fix, but I want to sit on it for a bit: #946

amadeus:
I've merged a fix for this into master

PR #946, opened 2017-05-10
Fix jsArrowFuncArgs matching bug

amadeus:
Fix for #945

This is a fix for ensuring that jsArrowFuncArgs only match if there's whitespace between the parens and `=>` symbol.

I feel pretty confident about this fix, but it's worth sitting on for a couple days to ensure it doesn't cause any regressions


bounceme:
did you see #942 ?

amadeus:
I did not, however this one I think is a bit better since it also handles tab and newline characters

bounceme:
arrows can't have a newline between the args and `=>`

amadeus:
Durp, my linter wasn't configured and didn't catch that when I messing with it.

PR #947, opened 2017-05-13
misc: go back to l:pos only when needed;

bounceme:


bounceme:
bracing for any fallout

PR #948, opened 2017-05-14
poc , slow , ugly, scope cleaning function factory

bounceme:


bounceme:
something like this for #947 might be better than leaving the dict-functions lying around in memory. the call time function creation is unfortunately slower, so this can't be merged as is. still interesting to implement

PR #949, opened 2017-05-15
silent function list

bounceme:


bounceme:
awful debug session just now, nevoid has some incompatible behavior with newlines

PR #959, opened 2017-05-19
cino-= and misc style changes: indent.vim

bounceme:


indent/javascript.vim
```diff
@@ -458,7 +459,7 @@ function GetJavascriptIndent()
   if l:line =~ '^[])}]\|^|}'
     return max([indent(num),0])
   elseif num
-    return indent(num) + s:W + switch_offset + bL + isOp
+    return max([indent(num) + get(l:,'case_offset',s:W) + switch_offset + bL + isOp, 0])
```
utix: I think case_offset visibility should be change as s`witch_offset`. Just set it line 431 to` s:sw()`.
Or better way, set it to `0` line 431 and line 440 to  `cino(`=`) -s:W` . So line 462  should be
`return max([indent(num) + s:W + switch_offset + +case_offset + bL + isOp, 0])`

bounceme: case_offset, if existing, replaces the &sw which is added for lines between `{}`. it isn't additional indent

bounceme: the switch_offset is also added

bounceme: i like to have the additional variable, `s:W` is used in other functions between the main return and the switch part

PR #960, opened 2017-05-20
Improve case indent, use cinoption '='

utix:
Was done in e6277212ed1818bf6349237c2be96949d5e217dd but behavior was lost


bounceme:
thanks, i made my own  implementation right now, which conforms to the other parts of the script a bit better, do you mind if i close this and put mine in #959 ?

utix:
@bounceme sure

PR #963, opened 2017-05-28
star comment find, as loop

bounceme:


bounceme:
could easily support `cino-star` with this as well, now 70 is the limit (default for `cino-star` )

PR #967, opened 2017-05-31
experimental: optimization with reserved words

bounceme:


bounceme:
this might be irrelevant to performance though

PR #973, opened 2017-06-04
completely implement cinoptions

bounceme:


bounceme:
not enough motivation right now

Issue #983, opened 2017-06-15
[flow] syntax issue for maybe types within parenthesis 


peey:
I'm having having syntax issues with the following:

```js
function parseFor(init: ?(N.VariableDeclaration | N.Expression)): N.ForStatement {
  return "nothing special";
}
```
vim highlights the second closing bracket as a syntax error

I've reduced this to the following:

```js
function parseFor(init: ?(N.Expression)): N.ForStatement {
  return "nothing special";
}
```

which reproduces the same error (but isn't very useful for demonstration of where this might be used)

And the following works fine without any errors: (checked for sanity testing)
```js
function parseFor(init: ?N.Expression): N.ForStatement {
  return "nothing special";
}

// or without the maybe type
function parseFor2(init: (N.VariableDeclaration | N.Expression)): N.ForStatement {
  return "nothing special";
}
```

peey:
Is this related to https://github.com/pangloss/vim-javascript/issues/901 ?

bounceme:
it's a problem with the parenthesized maybe type

dcosson:
I don't think this is the same as #901 but could be related. In that issue, syntax checking breaks later in the file after a certain character, in this case the closing paren is highlighted red and later things in the file syntax highlight ok. I constantly run into both issues, here is a simple case that reproduces this one:

![screen shot 2017-06-22 at 10 58 32 am](https://user-images.githubusercontent.com/708904/27448486-cc589c6a-5739-11e7-9761-6b38082b7628.png)

The combination of the last arg having an optional type and the function having a typed return value seems to be what breaks the highlighting.  In other permutations (no typed return value, or a non-optional arg type following the optional one) there is no problem. And as you can see, the line after the error is highlighted fine.

bounceme:
we need people contributing to the flow stuff, or , facebook could provide support themselves. they seem to want nuclide to remain first class above all else though

dcosson:
I took a brief look at it but I'm brand new to vim syntax so not fully following everything.  From what I can tell, it looks like at least one part of the problem is that if an arrow function specifies any return type, the parens get parsed as jsParens instead of jsFuncParens. From what i understand of the syntax it looks like it should be working though, haven't found the bug yet.

amadeus:
I've recently created a branch with some flow fixes, if you could do me a solid and test it: https://github.com/pangloss/vim-javascript/pull/991

PR #987, opened 2017-06-20
Start of ternary pattern is '?' followed by whitespace

pappasam:
Sort of fixes the following syntax highlighting issue by assuming ternary operator question mark is followed by whitespace:

https://github.com/pangloss/vim-javascript/issues/516

The ternary operator is the ultimate culprit here, so we should consider it when trying to find a solution for the following issue (which is not solved by this PR):

https://github.com/pangloss/vim-javascript/issues/983

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -162,7 +162,7 @@ syntax region  jsBlock                        matchgroup=jsBraces              s
 syntax region  jsModuleGroup        contained matchgroup=jsModuleBraces        start=/{/ end=/}/   contains=jsModuleKeyword,jsModuleComma,jsModuleAs,jsComment skipwhite skipempty nextgroup=jsFrom
 syntax region  jsSpreadExpression   contained matchgroup=jsSpreadOperator      start=/\.\.\./ end=/[,}\]]\@=/ contains=@jsExpression
 syntax region  jsRestExpression     contained matchgroup=jsRestOperator        start=/\.\.\./ end=/[,)]\@=/
-exe 'syntax region jsTernaryIf matchgroup=jsTernaryIfOperator start=/'.(get(g:,'javascript_plugin_flow') ? '\%(:\_s*\)\@<!' : '').'?/ end=/\%(:\|[\}]\@=\)/ contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression'
+exe 'syntax region jsTernaryIf matchgroup=jsTernaryIfOperator start=/'.(get(g:,'javascript_plugin_flow') ? '[^:=[:space:]]\@<=\_s*\zs' : '').'?/ end=/\%(:\|[\}]\@=\)/ contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression'
 
 syntax match   jsGenerator            contained /\*/ skipwhite skipempty nextgroup=jsFuncName,jsFuncArgs
 syntax match   jsFuncName             contained /\<[a-zA-Z_$][0-9a-zA-Z_$]*\>/ skipwhite skipempty nextgroup=jsFuncArgs,jsFlowFunctionGroup
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -162,7 +162,7 @@ syntax region  jsBlock                        matchgroup=jsBraces              s
 syntax region  jsModuleGroup        contained matchgroup=jsModuleBraces        start=/{/ end=/}/   contains=jsModuleKeyword,jsModuleComma,jsModuleAs,jsComment skipwhite skipempty nextgroup=jsFrom
 syntax region  jsSpreadExpression   contained matchgroup=jsSpreadOperator      start=/\.\.\./ end=/[,}\]]\@=/ contains=@jsExpression
 syntax region  jsRestExpression     contained matchgroup=jsRestOperator        start=/\.\.\./ end=/[,)]\@=/
-syntax region  jsTernaryIf                    matchgroup=jsTernaryIfOperator   start=/?/  end=/\%(:\|[\}]\@=\)/  contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression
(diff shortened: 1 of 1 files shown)


bounceme:
im not sure about disallowing spaceless ternary `?` to fix a issue with flow. what about `?` following `:` ?

syntax/javascript.vim
```diff
@@ -162,7 +162,7 @@ syntax region  jsBlock                        matchgroup=jsBraces              s
 syntax region  jsModuleGroup        contained matchgroup=jsModuleBraces        start=/{/ end=/}/   contains=jsModuleKeyword,jsModuleComma,jsModuleAs,jsComment skipwhite skipempty nextgroup=jsFrom
 syntax region  jsSpreadExpression   contained matchgroup=jsSpreadOperator      start=/\.\.\./ end=/[,}\]]\@=/ contains=@jsExpression
 syntax region  jsRestExpression     contained matchgroup=jsRestOperator        start=/\.\.\./ end=/[,)]\@=/
-syntax region  jsTernaryIf                    matchgroup=jsTernaryIfOperator   start=/?/  end=/\%(:\|[\}]\@=\)/  contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression
+syntax region  jsTernaryIf                    matchgroup=jsTernaryIfOperator   start=/?\_s/  end=/\%(:\|[\}]\@=\)/  contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression
```
bounceme: start=/[^:[:space:]]\_s*\zs?/

pappasam: Unfortunately, this suggestion does not work. It breaks the ternary operator's syntax highlighting. Sadly, since this plugin uses regex to parse language keywords, our expressiveness is limited.

With your suggestion:

image


With my suggestion:

image

pappasam:
image

The interruptions this causes for spaceless ternary expressions are minimal, albeit unfortunately present. The current state of the plugin completely destroys syntax highlighting in the presence of "Maybe" flowtypes. Completely understandable if you don't want to accept the tradeoffs of this pull request, but without it (or a better solution that actually works) this plugin is broken for those of us who use flow Maybe types.

pappasam:
To further clarify the problem, without this pull request:

image

With this pull request:

image

bounceme:
`start=/\%(:\_s*\)\@<!?/` does this work? it seems to for me.

sroecakepler:
@bounceme Does not work for me.

With start=/\%(:\_s*\)\@<!?/ in jsTernaryIf:

image

With start=/?\_s/:

image

bounceme:
is it the correct regex? the one you commented is different

sroecakepler:
sorry, didn't fence the code appropriately when I pasted

bounceme:
`start=/\%([:=]\_s*\)\@<!?/` ?

sroecakepler:
That works!

bounceme:
last thing: since `\@<!` is slow and the check is only needed for flow users

`exe 'syntax region jsTernaryIf matchgroup=jsTernaryIfOperator start=/'.(get(g:,'javascript_plugin_flow') ? '\%(:\_s*\)\@<!' : '').'?/  end=/\%(:\|[\}]\@=\)/ contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression'`

 still work?

bounceme:
@amadeus

bounceme:
https://github.com/pangloss/vim-javascript/pull/987#issuecomment-309824566

scratch that idea, the tradeoff is probably negligible.

bounceme:
does it work with this?
```
syntax region  jsTernaryIf                    matchgroup=jsTernaryIfOperator   start=/[^:[:space:]]\_s*?/lc=1  end=/\%(:\|[\}]\@=\)/  contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression
```

see `:h :syn-pattern-offset`

bounceme:
`start=/\%([^:=[:space:]]\@<=\_s*\zs\)?/` is working for me and should be faster. im not confident in my vim syntax rules though

pappasam:
@bounceme seems to work similarly

bounceme:
damn, these are causing huge slowdowns for me, even for indentation

bounceme:
another potential solution, with no performance drawback:
```
syntax region  jsTernaryIf                    matchgroup=jsTernaryIfOperator   start=/[^:[:space:]]\_s*?/lc=1  end=/\%(:\|[\}]\@=\)/  contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression
```

im not certain this actually works though

bounceme:
i give up

pappasam:
Yeah, unfortunately, the updates mentioned above bring us back to square one on for me.

bounceme:
i say @amadeus should make the call here

amadeus:
I'll need to look into this, however my personal stance is - Javascript syntax highlighting ALWAYS comes first. 

We are going to attempt to make this as accurate as possible for JS. If there's a problem with flow, we are not going to break Javascript syntax highlighting to fix it. In other words, enforcing a space after a `?` is an unacceptable solution imo.

I've just been crazy busy lately, which is why I haven't been very active here, I'll try and get back into the swing of things, perhaps this weekend to see if I can solve this in a better way.

pappasam:
@amadeus the current state of this pull request no longer requires a space after the ?. @bounceme updated the Ternary Regex to more-accurately reflect the requirements for a ternary operator and not conflict with flow syntax. At present, to the best of my knowledge, he's proposed a working solution. That said, the solution may have performance implications that I think we'd like you to consider. Thanks in advance for taking the time to consider this issue; flow syntax is pretty much broken without some sort of fix.

amadeus:
I think this issue should be fixed in #991

sroecakepler:
@bounceme after looking everything over, I agree that this specific issue is fixed in #991. Would you like to close this pull request?

PR #991, opened 2017-06-26
Misc Flow Fixes

### LINKED ISSUE #567: [flow] defining a function that returns another function

### LINKED ISSUE #723: Issue with with flow type object

### LINKED ISSUE #792: flow exact object
https://flowtype.org/docs/objects.html#exact-object-types

### LINKED ISSUE #845: class type parameters
```
class GenericClass<T> {
  x: T;
  constructor(x: T) {
    this.x = x;
  }
}
```
The `<T>` should have a flow syntax group applied

### LINKED ISSUE #885: [flow] import/export type should have syntax highlighting applied
Currently, type is highlighted when declared but export type is not highlighted.

`type foo` works
`export type foo` does not work

### LINKED ISSUE #891: numerous flowtype highlighting issues
My g:javascript_plugin_flow is set to 1, and the same issue exists while set to 0 (unsurprisingly).

If I create an optional type variable, say `type OptionalString = ?string`, the syntax highlighting below is broken.

If I create a functional type such as `type MyFunctionType = (val1: string) => string`, the syntax below is broken.

If I type a variable with `var1: Array<?string>`, the syntax below is broken.

### LINKED ISSUE #901: FlowType syntax breaks on nullable types
vim-javascript__flowtype_syntax_issue_1
vim-javascript__flowtype_syntax_issue_2

### LINKED ISSUE #966: [flow] Polymorphic classes
I also posted this issue on the [vim-flow](https://github.com/flowtype/vim-flow/issues/56) repo, but since then I saw https://github.com/pangloss/vim-javascript/issues/592 here which presents the possibility that it needs to be fixed for this plugin.

Using generics/polymorphism with es6 classes seems to break syntax highlighting. Here's the contents of the file causing the issue:

```javascript
// @flow

import { Component, Element } from 'react';
import { allDates } from './calendarData';
import range from './lib/range';

function selectedDate(value) {
  const date = (Date.parse(value) && new Date(value)) || new Date();
  date.setHours(0, 0, 0, 0);
  return date;
}

function startDate({ value }) {
  const date = selectedDate(value);

  return {
    month: date.getMonth() + 1,
    year: date.getFullYear()
  };
}

type Props = { value: string };
type State = { month: number, year: number };

export default class SlimProto<T> extends Component<void, Props & T, State> {
  state = startDate(this.props);

  componentWillReceiveProps(nextProps: Props) {
    if (this.props.value !== nextProps.value) {
      this.setState(startDate(nextProps));
    }
  }

  props: Props & T;

  previousYear = () => {
    this.setState(({ year }) => ({ year: year - 1 }));
  };

  nextYear = () => {
    this.setState(({ year }) => ({ year: year + 1 }));
  };

  previousMonth = () => {
    this.setState(({ month, year }) => {
      if (month === 1) {
        return { month: 12, year: year - 1 };
      }
      return { month: month - 1, year };
    });
  };

  nextMonth = () => {
    this.setState(({ month, year }) => {
      if (month === 12) {
        return { month: 1, year: year + 1 };
      }
      return { month: month + 1, year };
    });
  };

  selected() {
    return selectedDate(this.props.value);
  }

  rows() {
    const { month, year } = this.state;
    const { buckets } = allDates(month, year);
    const rowCount = Math.ceil(buckets.length / 7);

    return range(1, rowCount).map((n) => {
      const start = 7 * (n - 1);

      return {
        columns: buckets.slice(start, start + 7)
      };
    });
  }

  render(): ?Element<*> {
    return null;
  }
}
```

How it appears in my editor is inconsistent. The screenshot that I left on the vim-flow repo is what it looked like yesterday.  Today it appears as below. Besides the obvious red highlighted bracket, the `render` method also normally appears in blue.

image

### LINKED ISSUE #983: [flow] syntax issue for maybe types within parenthesis

amadeus:
Fixes #901
Fixes #966
Fixes #983
Fixes #891
Fixes #885
Fixes #845
Fixes #792
Fixes #723
Fixes #567

This is all a bit hairy, but I think I put a big dent in a lot of the common flow problems that have been reported. I want to sit on these changes for a few days though to make sure I didn't cause a serious regression.

### DIFF
--- a/extras/flow.vim
+++ b/extras/flow.vim
@@ -1,8 +1,8 @@
 syntax region  jsFlowDefinition     contained                        start=/:/    end=/\%(\s*[,=;)\n]\)\@=/ contains=@jsFlowCluster containedin=jsParen
 syntax region  jsFlowArgumentDef    contained                        start=/:/    end=/\%(\s*[,)]\|=>\@!\)\@=/ contains=@jsFlowCluster
-syntax region  jsFlowArray          contained matchgroup=jsFlowNoise start=/\[/   end=/\]/        contains=@jsFlowCluster
-syntax region  jsFlowObject         contained matchgroup=jsFlowNoise start=/{/    end=/}/         contains=@jsFlowCluster
-syntax region  jsFlowExactObject    contained matchgroup=jsFlowNoise start=/{|/   end=/|}/       contains=@jsFlowCluster
+syntax region  jsFlowArray          contained matchgroup=jsFlowNoise start=/\[/   end=/\]/        contains=@jsFlowCluster,jsComment
+syntax region  jsFlowObject         contained matchgroup=jsFlowNoise start=/{/    end=/}/         contains=@jsFlowCluster,jsComment
+syntax region  jsFlowExactObject    contained matchgroup=jsFlowNoise start=/{|/   end=/|}/       contains=@jsFlowCluster,jsComment
 syntax region  jsFlowParens         contained matchgroup=jsFlowNoise start=/(/  end=/)/ contains=@jsFlowCluster keepend
 syntax match   jsFlowNoise          contained /[:;,<>]/
 syntax keyword jsFlowType           contained boolean number string null void any mixed JSON array Function object array bool class
@@ -30,8 +30,8 @@ syntax match   jsFlowWildcardReturn contained /*/ skipwhite skipempty nextgroup=
 syntax region  jsFlowFunctionGroup  contained matchgroup=jsFlowNoise start=/</ end=/>/ contains=@jsFlowCluster skipwhite skipempty nextgroup=jsFuncArgs
 syntax region  jsFlowClassGroup     contained matchgroup=jsFlowNoise start=/</ end=/>/ contains=@jsFlowCluster skipwhite skipempty nextgroup=jsClassBlock
 
-syntax region  jsFlowTypeStatement                                   start=/type\%(\s\+\k\)\@=/    end=/=\@=/ contains=jsFlowTypeOperator skipwhite skipempty nextgroup=jsFlowTypeValue keepend
-syntax region  jsFlowTypeValue      contained                        start=/=/       end=/;/ contains=@jsFlowCluster,jsFlowGroup,jsFlowMaybe
+syntax region  jsFlowTypeStatement                                   start=/type\%(\s\+\k\)\@=/    end=/=\@=/ contains=jsFlowTypeOperator oneline skipwhite skipempty nextgroup=jsFlowTypeValue keepend
+syntax region  jsFlowTypeValue      contained     matchgroup=jsFlowNoise start=/=/       end=/;/ contains=@jsFlowCluster,jsFlowGroup,jsFlowMaybe
 syntax match   jsFlowTypeOperator   contained /=/ containedin=jsFlowTypeValue
 syntax match   jsFlowTypeOperator   contained /=/
 syntax keyword jsFlowTypeKeyword    contained type
--- a/extras/flow.vim
+++ b/extras/flow.vim
@@ -31,7 +31,7 @@ syntax region  jsFlowFunctionGroup  contained matchgroup=jsFlowNoise start=/</ e
 syntax region  jsFlowClassGroup     contained matchgroup=jsFlowNoise start=/</ end=/>/ contains=@jsFlowCluster skipwhite skipempty nextgroup=jsClassBlock
 
 syntax region  jsFlowTypeStatement                                   start=/type\%(\s\+\k\)\@=/    end=/=\@=/ contains=jsFlowTypeOperator oneline skipwhite skipempty nextgroup=jsFlowTypeValue keepend
-syntax region  jsFlowTypeValue      contained     matchgroup=jsFlowNoise start=/=/       end=/;/ contains=@jsFlowCluster,jsFlowGroup,jsFlowMaybe
+syntax region  jsFlowTypeValue      contained     matchgroup=jsFlowNoise start=/=/       end=/[\n;]/ contains=@jsFlowCluster,jsFlowGroup,jsFlowMaybe
 syntax match   jsFlowTypeOperator   contained /=/ containedin=jsFlowTypeValue
 syntax match   jsFlowTypeOperator   contained /=/
 syntax keyword jsFlowTypeKeyword    contained type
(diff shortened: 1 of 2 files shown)


pappasam:
This is amazing, thanks so much! So many fixes in one pull request. One noticeable bug remaining, though not sure if this is something worth addressing now, considering how much progress has been made:

image

Thanks again for spending time on flow issues, they're super important for my code base!

sebcode:
Thanks @amadeus! Few things:

flow

https://gist.github.com/sebcode/c691dbc1d3d00f541cfe80615026817d

I'll continue to work with that branch and see if I can find more cases.

pappasam:
@amadeus I found a fix for the function parameter coloring problem:

In extras/flow.vim, line 50, change:

```viml
syntax region  jsFlowParenAnnotation contained start=/:/ end=/)\@=/ containedin=jsParen contains=@jsFlowCluster
```

to include the options ```[=)]```:

```viml
syntax region  jsFlowParenAnnotation contained start=/:/ end=/[,=)]\@=/ containedin=jsParen contains=@jsFlowCluster
```

**Old coloring**

image


**New coloring**

image

I'd make the change myself, but I'm not a collaborator on this repository.

sebcode:
Added another testcase to the [gist](https://gist.github.com/sebcode/c691dbc1d3d00f541cfe80615026817d):

import

amadeus:
@sebcode thanks a ton for these code snippets!

So a couple things

1. Should be fixed - I added your change, which we should sit on for a bit to ensure no regressions
2. My code was working there, I just forgot to link the highlight group, so should be fixed
3. I think this was actually related to another issue I fixed the other day, TL;DR should be fixed.
4. This one may not really be possible to fix given the trying to regex match arrow functions. I'll add a follow-up comment that digs into it a bit more.
5. Similar issue to 2 - I just forgot to link a highlight group, should be fixed.

amadeus:
So on the subject of arrow functions, and matching arrow function args - it's very difficult to pull this off correctly, and honestly regexes are poorly equipped to detect them.

Part of the reason it's so difficult, is because we are attempting to match parenthesis that are then followed by an `=>`. On the surface this sounds pretty easy, but once you start to take in the possible edge cases, and how vim syntax systems work, it gets difficult very quickly.

Vim has a few different types of ways of creating matches, one of the more powerful ones is something called `region` The way you define a region is:

```
syntax region someName start=/(/ end=/)/ ...
```

Basically you define a regex to start and a regex to end the region. Regions then provide a sort of blank context, for you to put children regexes into. Think of them like a scope. One thing with regions though, the end doesn't need to match to start the regex, UNLESS you set the `oneline` option, which then means the region is limited to one line. Regions are also super powerful in that it makes it easy to nest other regions as well.

In the case of arrow function args, there's no easy way to parse them out vs a normal expression that contains parentheses until you get to that closing paren, and follow it up with a `=>`. Because of this, there are a bunch of limitations on how we match arrow function args, and it's very easy to break them.

Some example of how complex arrow function argument definitions can get, especially when you start introducing default params, destructurin, and even flow annotations

```js
// A simple example
() => true;
(test) =>{}
(test=1) => {}
(test=(2 + 4) * 3) => {}
({test}) => {}
({test=() => {}}) => {}
({test=() => {}}): () => {} => {}
```

Flow annotations add a whole other depth of complexity - since now the arrow operator does not follow the closing paren... So how can we build a regex that can both match the opening and closing parenthesis of the arrow function argument and still know that it's not some other parenthesis expression? Oh yeah, to make matters worse, it could be a multiline statement, which means that any regex you build for this could be inherently crazy expensive, even if you got it to work, probably making it not worth the while.

One proposal I posted to the Vim mailing list, was to add an optional feature to regions - that they must match the end regex in order to qualify. However Bram shot this down so I doubt it will get anywhere.

Honestly, the ultimate fix would be if we could pipe files through something like babel and get syntax groups based on that... then we could get rid of this syntax file completely xD, or at the very least just have it

sebcode:
I agree, syntax highlighting with regexes alone is error prone and probably impossible to get it all right. [Modern IDEs seem to use a mix of both](https://stackoverflow.com/questions/7243409/syntax-highlighting-how-does-eclipse-do-it-so-fast), basic syntax coloring and syntax coloring based on the parsed AST. Not sure if it's feasible to implement AST based syntax highlighting for vim. It's easy to generate the AST itself though: `flow ast test.js` (see also https://astexplorer.net/).

bounceme:
@sebcode too slow, requires valid ast, energy inefficient. you have conditions, skipping, predicting; vim's highlighting is adequate for most things.

amadeus:
It would need to be built into Vim as a possible feature. Hacking it in via a plugin would just not be possible I don't think :/

pappasam:
@amadeus the type statement's parsing (flow.vim, line 34) should be updated from:

```viml
syntax region  jsFlowTypeValue      contained     matchgroup=jsFlowNoise start=/=/       end=/;/ contains=@jsFlowCluster,jsFlowGroup,jsFlowMaybe
```

to

```viml
syntax region  jsFlowTypeValue      contained     matchgroup=jsFlowNoise start=/=/       end=/[\n;]/ contains=@jsFlowCluster,jsFlowGroup,jsFlowMaybe
```

Before:

image

After:

image

pappasam:
@amadeus @bounceme bump re: syntax update for extras/flow.vim, line 34

amadeus:
@pappasam Those tweaks should be in place, I will probably look to merge this soon unless there's any last minute things (this branch has been working rather well for me as of late).

PR #994, opened 2017-06-30
edge case fix: prevcodeline() and c style comments

bounceme:
since, `[*` is only `c` aware


bounceme:
this pr addresses
```c
/*
/*
 *
 */
```
edge case

PR #996, opened 2017-07-03
document and improve class cases

bounceme:


bounceme:
can remove the `extends {}` check since that will never work

Issue #998, opened 2017-07-05
[suggestion] indentation option `cino-W`


leomao:
I want to achieve the following indentation:
```javascript
// the last character is '('
a_long_function(
  arg1,
  arg2,
  arg3
);

// there are some arguments after '('
another_function(arg1, arg2,
                 arg3, arg4);
```
If I set `cino+=(2`, then in the second case, the second line will not be aligned with `(`.
If I set `cino+=(0`, then in the first case, `arg1` will be aligned with `(`.

In `cinoptions`, there is an option `cino-W`, then I can set `cino+=(0,W2` to achieve the desired indentation.  So I think we can also respect this option in the plugin.

PR #1000, opened 2017-07-05
cino-W

bounceme:
#998 this works?


leomao:
Yes, it works!  Thanks :)

PR #1015, opened 2017-08-04
cleaning, small optimization?

bounceme:


bounceme:
the "small" optimization can be fairly apparent, being in the bottleneck of the script

PR #1023, opened 2017-08-12
Add 'mjs' extension

brad-f:
Noticed this after experimenting with [es modules](https://blogs.windows.com/msedgedev/2017/08/10/es-modules-node-today/).


bounceme:
thanks!

PR #1031, opened 2017-08-23
experimental: reuse more code

bounceme:


bounceme:
probably a disorganized branch, for a while

bounceme:
timeout, which is useless with a search loop; stopline, args work exactly like search(), only appended an expression argument

PR #1033, opened 2017-08-25
Correctly highlight function bodies with an unnamed, generic function.

SamirTalwar:
Given the following JavaScript + Flow code:

    const foo = function<T>(x: T): void {
      // arbitrary, unrelated body
      while (true) {
        console.log(x);
      }
    }

The body won't be highlighted correctly; on my machine, it's mostly in red.

This seems to be caused by incorrectly parsing the `<T>` when the function is unnamed. Not sure why it ends up highlighting everything in one colour.

This commit fixes the problem. I might have missed somewhere; I looked for `jsFuncArgs` and added it to the same places.

I have not fixed the same issue for `const foo = <T>(x: T): void => …`. Because `<T>` marks the start of the arrow function, it's a bit more complicated and this is already past my comfort zone; it's my first foray into vim syntax-fu. Happy to look into it with a bit of guidance.


amadeus:
I think this should be safe! Thanks!

amadeus:
Arrow functions are super gnarly and no fun to try and match 😢

SamirTalwar:
Yeah, they look a bit terrifying, especially when Flow is involved. I'll leave it to a pro. Thanks for merging!

PR #1042, opened 2017-09-11
misc

bounceme:


utix:
@bounceme  I do a search using git bisect and I have found:

c217c783bd185d84fe7cee6d97b9b65e77ac3fe4 is the first bad commit

Since this commit I get
```
    bigquery.query({ 
                   query: query,
                   params: {                                                   
                       origin: origin                                          
                   }                                                           
    })
```

Whereas before I was getting

```
    bigquery.query({
        query: query,
        params: {
            origin: origin
        }   
    })  
```

Perhaps a cinoption can help.
When I finish a line with `{` I would like to get indent +s instead of be indent to the `{ `

PR #1053, opened 2017-09-27
adding optional chaining operator

npeeples:
https://github.com/tc39/proposal-optional-chaining
The only thing that could be bad about this is that 'x?.3:0' is
apparently allowed as a ternary operator. I'm not really sure if that
applies to the syntax highlighter, and is really not a great statement
anyway.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -36,7 +36,6 @@ syntax match   jsVariableDef    contained /\k\+/ skipwhite skipempty nextgroup=j
 syntax keyword jsOperator       delete instanceof typeof void new in of skipwhite skipempty nextgroup=@jsExpression
 syntax match   jsOperator       /[\!\|\&\+\-\<\>\=\%\/\*\~\^]\{1}/ skipwhite skipempty nextgroup=@jsExpression
 syntax match   jsOperator       /::/ skipwhite skipempty nextgroup=@jsExpression
-syntax match   jsOperator       /?\.\ze\_D/
 syntax keyword jsBooleanTrue    true
 syntax keyword jsBooleanFalse   false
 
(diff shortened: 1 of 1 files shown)


bounceme:
> (the ?. token must not be immediately followed by a decimal digit)

thanks for letting us know, will add the indentation parsing updates for this

bounceme:
here is the regex: `?\%(\.\d\@!\)\@!`

syntax/javascript.vim
```diff
@@ -162,7 +163,7 @@ syntax region  jsBlock                        matchgroup=jsBraces              s
 syntax region  jsModuleGroup        contained matchgroup=jsModuleBraces        start=/{/ end=/}/   contains=jsModuleKeyword,jsModuleComma,jsModuleAs,jsComment,jsFlowTypeKeyword skipwhite skipempty nextgroup=jsFrom fold
 syntax region  jsSpreadExpression   contained matchgroup=jsSpreadOperator      start=/\.\.\./ end=/[,}\]]\@=/ contains=@jsExpression
 syntax region  jsRestExpression     contained matchgroup=jsRestOperator        start=/\.\.\./ end=/[,)]\@=/
-syntax region  jsTernaryIf                    matchgroup=jsTernaryIfOperator   start=/?/  end=/\%(:\|[\}]\@=\)/  contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression
+syntax region  jsTernaryIf                    matchgroup=jsTernaryIfOperator   start=/?[^\.]/  end=/\%(:\|[\}]\@=\)/  contains=@jsExpression extend skipwhite skipempty nextgroup=@jsExpression
```
bounceme: will fail at eol

npeeples:
thanks for the feedback. not going to lie, I just threw some things in there to make it work, but I'm not really sure where that regex should go for a regular `?`. there doesn't seem to be anything that's just pulling out a regular question mark, and I'm not familiar enough with how the vim syntax files work to understand if there's something else that should be done.

npeeples:
Ok, I think I'm understanding now. That regex is for matching the start of the ternary operator... wow, that's quite the syntax for regex on these vim files.

bounceme:
I think @amadeus might be able to show us how it's done

PR #1056, opened 2017-09-28
[1029] - Vundle installation instructions

sam3k:
I would like to make a tiny addition to the README for those of us using Vundle but not quite as familiar with it.


bounceme:
https://github.com/pangloss/vim-javascript/commit/554ae9b24e7fb5bf7a6bd4ba8502758a9e18a744


thanks, but no thanks

PR #1059, opened 2017-10-06
(fix) Match no argument arrow functions

p-herbert:
# Description

Currently argument-less arrow function parentheses are not replaced with their concealing character.

# Solution

Fix `syntax match` statement.

# Results

### Before

```js
    const foo = () => // Do something
```

### After

```js
    const foo = λ => // Do something
```

Where `g:javascript_conceal_noarg_arrow_function='λ'`.


bounceme:
going to assume this works

PR #1060, opened 2017-10-11
clean: remove cursor call

bounceme:


bounceme:
if my hometown doesnt burn to the ground tonight, ill be able to finish this up

PR #1061, opened 2017-10-12
Changed recommended foldmethod setting in README

cjbassi:
Just a small change, thought I would make the foldmethod setting specific to javascript.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Default Value: 0
 -----------------
 
 ```vim
-augroup java_folding
+augroup javascript_folding
     au!
     au FileType javascript setlocal foldmethod=syntax
 augroup END
--- a/README.md
+++ b/README.md
@@ -52,13 +52,12 @@ Default Value: 0
 -----------------
 
 ```
-set foldmethod=syntax
+autocmd FileType javascript setlocal foldmethod=syntax
 ```
 
-Enables code folding based on our syntax file.
+Enables code folding for javascript based on our syntax file.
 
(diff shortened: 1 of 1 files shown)


bounceme:
thanks, i dont think the autocmd part is necessary though. how about just change `set` to `setlocal` ?

bounceme:
also:
```
'foldmethod' 'fdm'	string (default: "manual")
			local to window
```

does that mean it already is a local var?

bounceme:
hard to interpret `h local-options`

cjbassi:
After playing around with it a bit, it seems that `setlocal` sets the foldmethod for the current buffer, while `set` sets the foldmethod for the current buffer and all subsequently opened buffers. It would be better to do `setlocal` because for example you could open a javascript file and then a markdown file and the foldmethod setting gets carried over to the markdown file.

And then the `autocmd` part is so that `setlocal foldmethod=syntax` is called on every javascript file only and only when it's opened. So instead of putting `set foldmethod=syntax` in your vimrc where every while gets the setting, it's just javascript files.

bounceme:
it's good to put autocmds as members of an augroup.

cjbassi:
Oh yah you're right about the `au` group, I fixed that. Also true about the ftplugins folder. First time learning about that. At this point, I don't think it's that big of a deal how people implement the syntax folding, it probably won't cause any immediate issues, especially since you say in the README that it's a global option. Feel free to close this if you want.

PR #1074, opened 2017-11-10
split 2s timeout

bounceme:


bounceme:
I can't produce an example were this timeout ever is reached, even in a 100k line file

PR #1090, opened 2017-12-12
Fix jsFlowTypeKeyword

andreypopp:
Doesn't work without `containedin`.


andreypopp:
@amadeus makes sense, see #1091 which does that and also another fix

PR #1104, opened 2018-02-19
Highlight function calls same as function names 

texhnolyze:
I had the following issue: function calls on objects are not highlighted and are colored the same way as parens or "noise". I'd expect them to be the same color as function names (which is the handled like that in most editors/ides afaik).

I wasn't sure if this was a missing feature or a bug so I used this simple vim plugin config to reproduce  my issue securely:
```
call plug#begin('~/.config/nvim/plugged')

" javascript syntax highlighting and indentation
Plug 'pangloss/vim-javascript'

call plug#end()

" Enable syntax highlighting for JSDoc
let g:javascript_plugin_jsdoc = 1
```

Before:
no-fn-call-highlighting

After: 
fn-call-highlighting

I don't have a lot of knowledge of vimscript so if this should be addressed elswhere in the code could someone point me in that direction.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -285,6 +285,7 @@ if version >= 508 || !exists("did_javascript_syn_inits")
   HiLink jsGenerator            jsFunction
   HiLink jsArrowFuncArgs        jsFuncArgs
   HiLink jsFuncName             Function
+  HiLink jsFuncCall             Function
   HiLink jsClassFuncName        jsFuncName
   HiLink jsObjectFuncName       Function
   HiLink jsArguments            Special


amadeus:
This type of tweak is something users should do as a customization to their color scheme and not something that I think should be linked here. If a good number of people disagree than I am willing to merge this.

As a color scheme modification, you’d simply add a preferred color to setting to `jsFuncCall`

```
highlight jsFuncCall guifg=#ffffff
```

Just replace the guifg color with whatever you’d like to use for that color.

texhnolyze:
Ah ok thanks for the clarification. Should I keep this open for now to see if others agree or do you want me to close the pull request?

kristijanhusak:
I can agree on this. Most colorschemes that add `jsFuncCall`, just use the same value as `Function` highlighing, so setting that to default would solve this issue for other color schemes that doesn't have `jsFuncCall` defined.

Issue #1105, opened 2018-02-21
Syntax highlighting breaks when destructuring objects using computed properties


mikeycgto:
Consider the following JavaScript:

```js
function extract(type, obj) {
  const { [type]: { property } } = obj;

  return property;
}

// will return: 1
extract('foo', { foo: { property: 1 } });
```

#### Results

<img width="325" alt="screen shot 2018-02-21 at 11 06 55 am" src="https://user-images.githubusercontent.com/82063/36490887-5cb50690-16f7-11e8-8614-45d0ce3088da.png">

The `return` is not properly highlighted and code appearing after the destructuring fails to highlight correctly.

wzrdtales:
Same issue exists with

```javascript
const [ { some: [thing] }] = [ { some: [ '123' ] } ];
```

amadeus:
Thanks for the report, this issue should be fixed noe!

amadeus:
The second comment is a separate issue tho

wzrdtales:
you need me tp open a new one? Or will you do it by yourself?

amadeus:
Is fixed in master now

wzrdtales:
Thanks :) works.

PR #1109, opened 2018-03-02
Additional destructuring fix for objects within arrays

amadeus:
This should fix the second mentioned issue in #1105


wzrdtales:
ic, nvm then :D

PR #1127, opened 2018-05-27
Make a distinct group for keyword operators

chris-morgan:
In the stock Vim javascript syntax file, `new`, `delete`, `instanceof`
and `typeof` belong to syntax group `javaScriptOperator`, while symbol
operators have no syntax group. But this package was highlighting
keyword and symbol operators alike.

With this change, users can choose to return to highlighting keyword but
not symbol operators, with this in their `after/syntax/javascript.vim`:

    hi link jsOperatorKeyword Operator
    hi clear jsOperator


amadeus:
Thanks!

PR #1128, opened 2018-05-30
Highlight Flow opaque keyboard

jez:
Flow supports "opaque types", which are types whose definition is known in the
file where it's defined, but hidden in all other files (this allows creating
abstract types, a sort of information hiding.)

<https://flow.org/en/docs/types/opaque-types/>

As an example of the syntax:

```js
// @flow

// Type representing an opaque identifier.
// Outside of this file, Id can't be used as a number.
opaque type Id = number;

// Inside this file, Id *can* be used as a number.
let next: Id = 0;

export const freshId = (): Id => {
  return next++;
}
```

I've been using this change locally for a couple months without issue, so it
should be pretty safe to merge.


amadeus:
Looks good! Thanks!

PR #1136, opened 2018-07-14
Folds customization

zdm:
Following global variables added:

let g:javascript_fold_blocks = 1
let g:javascript_fold_comments = 1
let g:javascript_fold_functions = 1

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ Enables code folding for javascript based on our syntax file.
 
 Please note this can have a dramatic effect on performance.
 
+You can choose, what to fold: 
+```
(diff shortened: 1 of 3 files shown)


zdm:
This PR allows users to choose, what to fold: comments, functions, other code blocks.

vokar:
Any possible progress regarding merging this PR?

amadeus:
Unfortunately I am personally opposed to merging this - I don't really like converting a ton of syntax definitions to exec blocks, and we lose syntax highlighting.

zdm:
I agree, that syntax is pretty ugly.
But is there is another way to fold functions only?

PR #1058, opened 2017-09-29
Added more operator conceal

chinleung:
Allow the following conceal options:

```
let g:javascript_conceal_double_equal = "≈"
let g:javascript_conceal_triple_equal = "≡"
let g:javascript_conceal_or_operator = "∨"
let g:javascript_conceal_and_operator = "∧"
let g:javascript_conceal_not_operator = "¬"
let g:javascript_conceal_greater_equal_operator = "≥"
let g:javascript_conceal_less_equal_operator = "≤"
let g:javascript_conceal_not_equal_operator = "≠"
let g:javascript_conceal_not_strict_equal_operator = "≢"
```

PR #1140, opened 2018-08-20
"Added more operator conceal"

antialias:
cherry-picked relevant commits from https://github.com/pangloss/vim-javascript/pull/1058 and correctly based off `pangloss/master`


amadeus:
Had to revert these commits since this completely broke detection of a lot of things (like `!` or `||`, etc). I don't really have time to debug why though, so if you guys still want this you'll need to figure out how to fix all the broken items.

PR #1141, opened 2018-08-29
Fixes #1123

### LINKED ISSUE #1123: String class fields mess up highlighting of when omitting semicolons
*Requisite minimal reproducible example, formatted as plain text :*

    class Faulty {
      foo = 'bar'
      static doesNotWork1 = 'bar'
      static doesNotWork2 = "bar"
      static x
    }

In the class above, static is never highlighted.

<hr>

String properties without a trailing semicolon messes up the coloring of the next line.

adrianhelvik:


amadeus:
Interesting! Thanks for the fix!

PR #1161, opened 2019-02-19
Improve Generic Support

amadeus:
The biggest feature upgrade here is actual support for generics in function calls:

```
arr.reduce<number>(callback, 0);
```

Additionally I took a chance to improve other areas of generic support, as well as fix a bunch of other random bugs I found that are much too annoying to document here.

It's possible these changes may have other unintended consequences, so I will sit on this PR for a few days before merging.


TrySound:
Awesome! Thank you @amadeus

Issue #1174, opened 2019-06-28
Laggy indent performance on large objects


amadeus:
I've attached a gist to use as an example

https://gist.github.com/amadeus/139a4a8d711a18444b0699bdea2ea27c

If you put your cursor on line 6001, and press `o` to jump into insert mode on a new line, it will be a bit laggy. If you press enter a few times while in insert mode, you get even more lag.

Is there a way we can optimize this use case @bounceme?

bounceme:
i'm looking into it! this definetly needs a fix

bounceme:
so the offending function is here: https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L293

the reason so far as i can tell is that it cant find the opening brace of the object before the timeout here:
https://github.com/pangloss/vim-javascript/blob/master/indent/javascript.vim#L293
 (the 2000 : 200 part)

so it wants to find the start of a continued expression and runs isContOne like 6000 times

bounceme:
so basically just need to limit the looping in that function. will fix soon

bounceme:
also limiting the looping will prevent the time lag, but it will probably make the indent incorrect, which is a tradeoff well worth it in these unusual spots

amadeus:
I know nothing about how indentation works - but is it possible it could just grab the syntax group it's in - and just match from there?

I.E. if it's in a `jsObject` and the previous line ends on a column - match the current indent or something?

bounceme:
you just gave me an idea.... basically a short circuit to get indent in objects really easily. i'll be working on this

PR #1175, opened 2019-07-02
laggy indent in object

bounceme:


bounceme:
https://github.com/pangloss/vim-javascript/issues/1174

this should fix it! thanks for the idea!

amadeus:
Nice! While on the subject of this - should we also include `jsBracket` (for arrays), `jsDestructuringBraces` and `jsClassBlock`?

amadeus:
Also, this definitely fixes the test case I created - however, it doesn't appear to fix the issue where I initially found it in a special file related to my work. I can't share that file - but it's a massive object with a lot of strings for translation. I'll do a quick profile and post here what seems to get lagged up in this particular case

amadeus:
This gist is a profile I ran, at the bottom you can see what functions seem to get called a TON:

https://gist.github.com/amadeus/c5f518a008ee1824dbff2d86e4610a36

The syntax group I am in is `jsParen > jsObject `

bounceme:
how are u generating these profiling logs? i havent vimscript'd in a long time now...
yea, there are many bottlenecks in giant list/objects. i have a few more of these to address with this pr.

amadeus:
This is the vimscript I use to profile:

```
function! s:ProfileStart()
  profile start profile.log
  profile func *
  profile file *
endfunction

function! s:ProfileEnd()
  profile pause
  noautocmd qall!
endfunction

command! ProfileStart call s:ProfileStart()
command! ProfileEnd call s:ProfileEnd()
```

Basically I setup the situation (open the file move my cursor to the right location, then execute `:ProfileStart` then I perform the actions - then call `:ProfileEnd`. This stops the profile and creates a `profile.log` file in the current working directory of vim at the time of the profile.

Issue #1177, opened 2019-07-18
Syntax highlighting broken after a template string


yuezk:
The code syntax after a template string was highlighted as a string incorrectly. Below is the snapshot:

![image](https://user-images.githubusercontent.com/3297602/61473093-4cdae000-a9b8-11e9-99ae-23fbe00a367b.png)

The minimal vim config

```vim
call plug#begin('~/.vim/plugged')

Plug 'pangloss/vim-javascript'

call plug#end()

filetype plugin on
filetype indent on
```

The test code:

```js
const myVariable = "hello world";
const checkHightlighting = (({
  firstVariable,
  secondVariable,
  thirdVariable,
  fourthVariable,
  fifthVariable = "createdObject",
  sixthVariable = `create${myVariable}`
}) => {
  return verify => {
    const noHighlightingHere = true;
    return verify(noHighlightingHere);
  };
},
[myVariable]);
```

FabioAntunes:
If it helps here's another example that breaks syntax highlighting:

```javascript
Something.prototype.generaterOutput = function(data) {
    const text = `{1:F01${this.getText1()}0001000001}{2:I${this.getText2()}${this.getText3(data)}N}{3:${this.getText4()}}{4:${this.getText5()}\r\n-}`;

    return text;
};
```

richchurcher:
Another example, using the latest version (via vim-polyglot):
```
  getEditURL = communication => `/communications/${this.props.contract.id}/${communication.id}`
```
![works](https://user-images.githubusercontent.com/171905/74616173-3c2b5c00-518b-11ea-9581-7cbe40f2bc55.png)

Works fine without the newline, but as formatted by Prettier:
```
  getEditURL = communication =>
    `/communications/${this.props.contract.id}/${communication.id}`
```
![broken](https://user-images.githubusercontent.com/171905/74616187-44839700-518b-11ea-8826-04f92feafd36.png)

breaks! Would love to contribute a fix, but Vim highlighting has always been a bit of a mystery to me :confused:

ArthurJahn:
latest version works fine for me. See if you are using the plugin syntax highlight with the command:
```sh
:echo filter(split(execute(':scriptname'), "\n"), 'v:val =~? "javascript"')
```
The result might be like:
```sh
[...]

PR #1184, opened 2019-08-18
Fix highlighting of template strings within objects (#1177)

luizribeiro:
Template strings aren't matching correctly when used within object (see #1177 for an example). This fixes that by allowing the `jsObject` group to contain `jsTemplateString`.


luizribeiro:
Can we get this merged in by any chance?

amadeus:
I have a lot of mixed feelings on this - it's technically wrong from a syntax highlighting perspective - because the syntax for default args on a destructuring statement are not valid object syntax. But on the flip side - it's nearly impossible for us to properly highlight those statements due to their complexity. So I am willing to give this a try since I think it does prevent things from being broken

luizribeiro:
Got it. I'm not an expert on vim's syntax highlighting to be honest. If this causes trouble, feel free to revert. Thanks for taking a look!

PR #1198, opened 2020-01-24
feat: Add `NOTE` to `jsCommentTodo`

corbinhesse:
Highlight `NOTE` to mention in documentation or JSDoc

Additional comment flags from `rake notes` are seldom used and could pertain to the existing flags: OPTIMIZE/HACK/BUG works with FIXME or XXX. `NOTE` is needed to provide additional insight into comments for larger teams.


amadeus:
Thanks for the contribution!

PR #1202, opened 2020-04-06
Fix matching of jsClassProperty without an initial value

ohcibi:
Given the following code

```javascript
class Foo {
  bar;

  baz = 'biz';
}
```

The `baz` is properly matched as a `jsClassProperty` but `bar is
`jsClassBody`. The regex for `jsClassProperty` was requiring the `=`
that followed the property name. The changes allow value less
properties by allowing either `;` or `=` to follow the name.

`jsClassPropertyComputed` was not affected since that is matched by using the delimiters `[]` and
that works regardless of an initial value. But I noticed that for properties that have no semicolons
then the matching does not work for property values as well. I don't know if the recent spec
actually requires semicolons, so I didn't add it yet. But I would add it if class properties without
semicolons are still allowed.


amadeus:
Thanks for the PR!

PR #1205, opened 2020-05-03
add support for proposal-partial-application operator

morr:
https://babeljs.io/docs/en/babel-plugin-proposal-partial-application

Before:
image

After:
image


vieira:
It seems it would not work for this case?

```js
1 + 1
  |> f(1, ?)
```

vieira:
This seems to work for both cases and does not highlight the following `,` or `)` as part of the `?`.

```js
syntax match jsFuncArgPartApp contained /\%([(,]\s*\)\@<=?\%(\s*[,)]\)\@=/ containedin=jsFuncArgs
```

PR #1215, opened 2020-07-24
Add known DOM element properties

jsit:
From MDN: https://developer.mozilla.org/en-US/docs/Web/API/Element#Properties

Not sure if this is the best way to do this. But, for instance, if you are doing something like this in HTML:

```html
<div onclick="classList.toggle('foobar');"></div>
```

the `classList` property there doesn't get properly highlighted, since it doesn't follow a period. Might be useful to be explicit about these, although it also might be smart to limit the context in which they're valid matches.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -30,6 +30,13 @@ syntax match   jsObjectProp     contained /\<\K\k*/
 syntax match   jsFuncCall       /\<\K\k*\ze\s*(/
 syntax match   jsParensError    /[)}\]]/
 
+" Known Properties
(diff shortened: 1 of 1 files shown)


jsit:
Ah, I'm seeing @amadeus actually explicitly doesn't want this stuff in here:

https://github.com/pangloss/vim-javascript/commit/6c2de10

Not that this is anybody's problem but my own, but out of curiosity, if this shouldn't go here, where should it go? Surely not in the HTML syntax. In a separate syntax file altogether?

jsit:
Would it be overzealous to define as `jsObject` any string that's followed by a period?

```vim
syntax match jsObject /\<\K\k*\ze\s*\./
HiLink jsObject Identifier
```

PR #1218, opened 2020-08-09
Fix typescript detection bug

parmort:
I realized that when in a typescript file that includes a shebang line, the filetype is incorrectly detected as javascript. This check stops this incorrect detection.

My first thought was to add a configuration variable (hence the branch name), but I realized how limiting that is.


rschristian:
Just ran into this myself tonight, was scratching my head for a while over potential syntax issues before the shebang dawned on me as potentially being problematic (then took me a while longer to figure out what was mistakenly detecting it). 

Would be nice to get this in.

PR #1230, opened 2021-03-09
Fixes highlighting for finally in certain cases.

### LINKED ISSUE #1222: Try-catch-finally not properly highlighted
This valid JavaScript snippet is not highlighted properly in that `finally` is not considered a keyword like the former two are:

```js
try {} catch {} finally {}
```

This one instead is highlighted correctly:

```js
try {} catch() {} finally {}
```

thomasstep:
`jsCatch` can be followed by `jsParenCatch` or `jsTryCatchBlock`. A `catch` can either be followed directly by a block or [optionally have parens for an error parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#syntax). The original issue noticed that in the code
```js
try {} catch {} finally {}
```
`finally` was not realized as a keyword. This PR fixes that by allowing the `nextgroup` to be either parens or a block.


thomasstep:
Fixes #1222

amadeus:
Hah - i totally fixed this locally and forgot to push it, thanks!

PR #1238, opened 2021-05-20
Add WeakRef

vieira:
Add syntax highlightling to WeakRef as defined in [tc39/proposal-weakrefs](https://github.com/tc39/proposal-weakrefs).


amadeus:
Thanks!

PR #1242, opened 2021-09-24
Add missing lisp option to b:undo_indent

dkearns:


amadeus:
I don't have a lot of context on this, what does this fix?

dkearns:
You're setting the `'lisp'` option and not restoring it via the `b:undo_indent` variable as you are for all other options set in the plugin.

See `:help undo_indent`

amadeus:
thanks for the context!

PR #1244, opened 2021-11-04
Allow newlines between function name and parens for function call

hallzy:
The intent is to have syntax highlighting for function calls where the paren is on a new line. This is the style guide for my work when the arguments list is big and/or contains functions so having highlighting for it would be nice.

example:

```javascript
myFunc
(
    () =>
    {
        // Some callback
    }
);
```


amadeus:
Sorry for the delays, thanks!

PR #1249, opened 2022-07-15
Better distinguish statements from expressions

### LINKED ISSUE #1228: Code incorrectly highlighted in nested braces
IMG_20210122_205937
`let b=2;` is expected to be highlighted as a statement rather than a JSON syntax.

lassipulkkinen:
Restricts usage of `jsAll` to contexts where statements are allowed. This makes it possible to add `jsBlock` to `jsAll` (fixes #1228), which is also done here.

This also splits `jsRepeatParen` into `jsWhileParen` and `jsForParen`, since the syntaxes are obviously very different, and restricts `of` to `jsForParen`, since that's the only place where it actually has special meaning (this is still a bit crude, as it doesn't differentiate for-of from for...). Somehow [miraculously](https://github.com/pangloss/vim-javascript/commit/a0d1abc172cd0d701c115b7c0197c5dfa288f278) there was already a highlighting rule for `jsOf`, so that didn't need to be added.

### DIFF
--- a/syntax/javascript.vim
+++ b/syntax/javascript.vim
@@ -24,7 +24,8 @@ syntax sync fromstart
 " syntax case ignore
 syntax case match
 
-syntax match   jsNoise          /[:,;]/
+syntax match   jsNoise          /[:;]/
+syntax match   jsNoise          /,/ skipwhite skipempty nextgroup=@jsExpression
 syntax match   jsDot            /\./ skipwhite skipempty nextgroup=jsObjectProp,jsFuncCall,jsPrototype,jsTaggedTemplate
 syntax match   jsObjectProp     contained /\<\K\k*/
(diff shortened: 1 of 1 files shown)


amadeus:
I don't write much JS these days anymore (mostly typescript, and therefore a different plugin altogether).  That said though, it looks like you put some good effort into this. I'll go ahead and merge it, although I don't really have time to test it, so I'll keep an eye out and may revert if there are issues, or if you could also keep an eye out and submit any follow ups if there are issues, that would be helpful as well!

lassipulkkinen:
Thanks! I've been using this for the past few weeks and haven't encountered any issues, although obviously I did just force-push to fix a dumb typo, so there could be more..?

I won't necessarily be writing much JS in the near future either, though, as I kind of ended up abandoning the project I was working on when I submited this.

Issue #1257, opened 2025-03-12
freeze on paste with foldmethod=syntax


jayk:
I have had this issue for some time and only just tracked down the config that causes it.

When I have vim-javascript enabled and `set foldmethod=syntax` enabled, any attempt to paste will freeze vim. 

If I hit ctrl-c I can usually break out of it, but only part of the paste is present.

I thought it was an interaction with another plugin, but it turns out it just doesn't seem to like `foldmethod=syntax`

I've attached a minimal vimrc that triggers the problem.

**What I did:**  Used the attached vimrc, Opened a javascript file, navigated to a line inside the existing javascript, hit `i` and attempted to paste some javascript into it.

**What I expected:** Javascript would be pasted into the file at the location.

**What happened instead:** Vim hangs until I hit ctrl-c.

Vim version info:
```
VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Oct 06 2024 20:33:49)
Included patches: 1-764
Compiled by Arch Linux
Huge version without GUI.  Features included (+) or not (-):
+acl               +file_in_path      +mouse_urxvt       -tag_any_white
+arabic            +find_in_path      +mouse_xterm       +tcl/dyn
+autocmd           +float             +multi_byte        +termguicolors
+autochdir         +folding           +multi_lang        +terminal
-autoservername    -footer            -mzscheme          +terminfo
-balloon_eval      +fork()            +netbeans_intg     +termresponse
+balloon_eval_term +gettext           +num64             +textobjects
-browse            -hangul_input      +packages          +textprop
++builtin_terms    +iconv             +path_extra        +timers
+byte_offset       +insert_expand     +perl/dyn          +title
+channel           +ipv6              +persistent_undo   -toolbar
+cindent           +job               +popupwin          +user_commands
-clientserver      +jumplist          +postscript        +vartabs
-clipboard         +keymap            +printer           +vertsplit
+cmdline_compl     +lambda            +profile           +vim9script
+cmdline_hist      +langmap           -python            +viminfo
+cmdline_info      +libcall           +python3/dyn       +virtualedit
+comments          +linebreak         +quickfix          +visual
+conceal           +lispindent        +reltime           +visualextra
+cryptv            +listcmds          +rightleft         +vreplace
+cscope            +localmap          +ruby/dyn          +wildignore
+cursorbind        +lua/dyn           +scrollbind        +wildmenu
+cursorshape       +menu              +signs             +windows
[...]

PR #1258, opened 2025-03-21
Revise README to include a warning about foldmethod=syntax hanging du…

jayk:
This clarifies in the appropriate section of the README.md that foldmethod=syntax can cause hanging when pasting in some terminals, to hopefully help future travellers.  Re: issue #1257


amadeus:
Thanks for the PR!

